Issue #5, opened 2018-12-13
Simpler Storage API


DisposaBoy:
It's probably too late to change the API now, so this is more of a question about why the API is designed the way it is than anything.

## What would you like to have changed?

Remove `Locker` from the `Storage` interface.

## Why is this feature a useful, necessary, and/or important addition to this project?

`TryLock` is hard to implement correctly and the implementation doesn't seem to make use of it, because it just waits anyway.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

The ideal interface would be `Storage` on its own without `Locker`, then all methods would be required to be safe for concurrent use.
Apart from being more idiomatic, it makes things a lot simpler because the actual storage implementation has a better idea about how to do locking than CertMagic.

The main motivation was to implement a `Storage` in SQLite (or other databases) but I was blocked by the requirement that `TryLock` being non-blocking.
Instead of implementing the equivalent of `mu.Lock()`, it seemed I'd have to introduce a whole bunch of state that will probably be buggy for unknown reasons. Even `FileStorage.TryLock` makes me nervous because at a *quick glance*, it doesn't seem to handle cleanup so if the process crashes, everything is *probably* borked.

mholt:
This is a good question, but I am still convinced the two interfaces are jointly required for a workable Storage implementation.

It's not the Storage methods that need to be made safe for concurrent use -- they already need to be, of course -- it's that the ACME operations like the obtaining of a certificate need to be synchronized. This means not only one operation at a time, but one _in total_. TryLock is the easiest way I could think of that allows other processes to wait while another finishes obtaining the certificate it needs, then simply loading that certificate without obtaining it again. This idempotency is crucial for running correctly in a cluster, and it's why other libraries such as autocert don't do well sharing storage imementations even though it has a Cache interface.

Look closer at CertMagic's code and you'll see that it doesn't always wait -- it only waits if the lock is already held, and then it does some other flow when the waiting is over, whereas with a regular mutex you just wait and then continue, but here we have to alter the program flow.

For database backed storage you can use a transaction to "obtain" a lock or, if already obtained, to know that and simply return a waiter instead.

I hope that makes sense. They used to be separate but I really could not justify totally separating the two because it was not safe to use one but not the other.

mholt:
[...]

PR #7, opened 2018-12-13
Replace TryLock and Wait with Lock, and check for idempotency

mholt:
This replaces TryLock and Wait with a single Lock function, and we check after obtaining a lock to make sure Obtain and Renew remain idempotent.

See issue #5.

/cc @DisposaBoy - please help test and review!

### DIFF
--- a/filestorage.go
+++ b/filestorage.go
@@ -31,6 +31,8 @@ import (
 // cross-platform way or persisting ACME assets on the file system.
 type FileStorage struct {
 	Path string
+
+	fileStorageNameLocks map[string]*fileStorageWaiter
 }
 
 // Exists returns true if key exists in fs.
@@ -149,18 +151,26 @@ func dataDir() string {
 	return filepath.Join(baseDir, "certmagic")
 }
 
-// TryLock attempts to get a lock for name, otherwise it returns
-// a Waiter value to wait until the other process is finished.
-func (fs FileStorage) TryLock(key string) (Waiter, error) {
+// Lock obtains a lock named by the given key. It blocks
+// until the lock can be obtained or an error is returned.
+func (fs FileStorage) Lock(key string) error {
+	// can't defer the unlock because we might have
+	// to Wait() for a while before returning, so we're
+	// careful to unlock at all the right places
 	fileStorageNameLocksMu.Lock()
-	defer fileStorageNameLocksMu.Unlock()
+
+	if fs.fileStorageNameLocks == nil {
+		fs.fileStorageNameLocks = make(map[string]*fileStorageWaiter)
+	}
 
 	// see if lock already exists within this process - allows
 	// for faster unlocking since we don't have to poll the disk
-	fw, ok := fileStorageNameLocks[key]
+	fw, ok := fs.fileStorageNameLocks[key]
 	if ok {
 		// lock already created within process, let caller wait on it
-		return fw, nil
+		fileStorageNameLocksMu.Unlock()
+		fw.Wait()
+		return nil
 	}
 
 	// attempt to persist lock to disk by creating lock file
@@ -168,7 +178,8 @@ func (fs FileStorage) TryLock(key string) (Waiter, error) {
 	// parent dir must exist
 	lockDir := fs.lockDir()
 	if err := os.MkdirAll(lockDir, 0700); err != nil {
-		return nil, err
+		fileStorageNameLocksMu.Unlock()
+		return err
 	}
 
 	fw = &fileStorageWaiter{
@@ -199,19 +210,23 @@ createLock:
 			}
 
 			// if lock is not stale, wait upon it
-			return fw, nil
+			fileStorageNameLocksMu.Unlock()
+			fw.Wait()
+			return nil
 		}
 
 		// otherwise, this was some unexpected error
-		return nil, err
+		fileStorageNameLocksMu.Unlock()
+		return err
 	}
 	lf.Close()
 
 	// looks like we get the lock
 	fw.wg.Add(1)
-	fileStorageNameLocks[key] = fw
+	fs.fileStorageNameLocks[key] = fw
+	fileStorageNameLocksMu.Unlock()
 
-	return nil, nil
+	return nil
 }
 
 // Unlock releases the lock for name.
@@ -219,7 +234,11 @@ func (fs FileStorage) Unlock(key string) error {
 	fileStorageNameLocksMu.Lock()
 	defer fileStorageNameLocksMu.Unlock()
 
-	fw, ok := fileStorageNameLocks[key]
+	if fs.fileStorageNameLocks == nil {
+		fs.fileStorageNameLocks = make(map[string]*fileStorageWaiter)
+	}
+
+	fw, ok := fs.fileStorageNameLocks[key]
 	if !ok {
 		return fmt.Errorf("FileStorage: no lock to release for %s", key)
 	}
@@ -239,15 +258,18 @@ func (fs FileStorage) Unlock(key string) error {
 
 	// clean up in memory
 	fw.wg.Done()
-	delete(fileStorageNameLocks, key)
+	delete(fs.fileStorageNameLocks, key)
 
 	return nil
 }
 
-// UnlockAllObtained removes all locks obtained by
-// this instance of fs.
+// UnlockAllObtained removes all locks obtained
+// by this instance of fs.
 func (fs FileStorage) UnlockAllObtained() {
-	for key, fw := range fileStorageNameLocks {
+	if fs.fileStorageNameLocks == nil {
+		fs.fileStorageNameLocks = make(map[string]*fileStorageWaiter)
+	}
+	for key, fw := range fs.fileStorageNameLocks {
 		err := fs.Unlock(fw.key)
 		if err != nil {
 			log.Printf("[ERROR][%s] Releasing obtained lock for %s: %v", fs, key, err)
@@ -284,11 +306,11 @@ type fileStorageWaiter struct {
 	wg       *sync.WaitGroup
 }
 
-// Wait waits until the lock is released.
+// Wait waits until the lock at fw.filename is released.
 func (fw *fileStorageWaiter) Wait() {
 	start := time.Now()
 	fw.wg.Wait()
-	for time.Since(start) < 1*time.Hour {
+	for time.Since(start) < staleLockDuration {
 		_, err := os.Stat(fw.filename)
 		if os.IsNotExist(err) {
 			return
@@ -297,12 +319,12 @@ func (fw *fileStorageWaiter) Wait() {
 	}
 }
 
-var fileStorageNameLocks = make(map[string]*fileStorageWaiter)
-var fileStorageNameLocksMu sync.Mutex
-
 var _ Storage = FileStorage{}
-var _ Waiter = &fileStorageWaiter{}
 
 // staleLockDuration is the length of time
 // before considering a lock to be stale.
 const staleLockDuration = 2 * time.Hour
+
+// fileStorageNameLocksMu guard access to all
+// FileStorage.fileStorageNameLock values.
+var fileStorageNameLocksMu sync.Mutex
--- a/storage.go
+++ b/storage.go
@@ -64,22 +64,24 @@ type Storage interface {
 // Locker facilitates synchronization of certificate tasks across
 // machines and networks.
 type Locker interface {
-	// TryLock will attempt to acquire the lock for key. If a
-	// lock could be obtained, nil values are returned as no
-	// waiting is required. If not (meaning another process is
-	// already working on key), a Waiter value will be returned,
-	// upon which you should Wait() until it is finished.
+	// Lock acquires the lock for key, blocking until the lock
+	// can be obtained or an error is returned. Note that, even
+	// after acquiring a lock, an idempotent operation may have
+	// already been performed by another process that acquired
+	// the lock before - so always check to make sure idempotent
+	// operations still need to be performed after acquiring the
+	// lock.
 	//
 	// The actual implementation of obtaining of a lock must be
-	// an atomic operation so that multiple TryLock calls at the
+	// an atomic operation so that multiple Lock calls at the
 	// same time always results in only one caller receiving the
-	// lock. TryLock always returns without waiting.
+	// lock at any given time.
 	//
 	// To prevent deadlocks, all implementations (where this concern
 	// is relevant) should put a reasonable expiration on the lock in
 	// case Unlock is unable to be called due to some sort of network
 	// or system failure or crash.
-	TryLock(key string) (Waiter, error)
+	Lock(key string) error
 
 	// Unlock releases the lock for key. This method must ONLY be
 	// called after a successful call to TryLock where no Waiter was
@@ -98,11 +100,6 @@ type Locker interface {
 	// should be printed or logged, since there could be multiple,
 	// with no good way to handle them anyway.
 	UnlockAllObtained()
-}
-
-// Waiter is a type that can block until a lock is released.
-type Waiter interface {
-	Wait()
 }
 
 // KeyInfo holds information about a key in storage.
--- a/certificates.go
+++ b/certificates.go
@@ -64,12 +64,11 @@ func (c Certificate) NeedsRenewal() bool {
 	if c.NotAfter.IsZero() {
 		return false
 	}
-	timeLeft := c.NotAfter.UTC().Sub(time.Now().UTC())
 	renewDurationBefore := DefaultRenewDurationBefore
 	if len(c.configs) > 0 && c.configs[0].RenewDurationBefore > 0 {
 		renewDurationBefore = c.configs[0].RenewDurationBefore
 	}
-	return timeLeft < renewDurationBefore
+	return time.Until(c.NotAfter) < renewDurationBefore
 }
 
 // CacheManagedCertificate loads the certificate for domain into the
--- a/client.go
+++ b/client.go
@@ -208,6 +208,8 @@ func (cfg *Config) newACMEClient(interactive bool) (*acmeClient, error) {
 	return c, nil
 }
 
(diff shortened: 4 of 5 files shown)


filestorage.go
```diff
@@ -149,33 +151,46 @@ func dataDir() string {
...
+// until the lock can be obtained or an error is returned.
+func (fs FileStorage) Lock(key string) error {
+	// can't defer the unlock because we might have
+	// to Wait() for a while before returning, so we're
+	// careful to unlock at all the right places
 	fileStorageNameLocksMu.Lock()
-	defer fileStorageNameLocksMu.Unlock()
+
+	if fs.fileStorageNameLocks == nil {
+		fs.fileStorageNameLocks = make(map[string]*fileStorageWaiter)
```
DisposaBoy: `fs` is a `copy` so this assignment has no external effect.

I fixed it in https://github.com/DisposaBoy/certmagic/commit/36c74006a73e70a707b23c35828dd146f3b3c4ea so I could continue testing.

filestorage.go
```diff
@@ -149,33 +151,46 @@ func dataDir() string {
...
+		fileStorageNameLocksMu.Unlock()
+		fw.Wait()
+		return nil
 	}
 
 	// attempt to persist lock to disk by creating lock file
 
 	// parent dir must exist
 	lockDir := fs.lockDir()
 	if err := os.MkdirAll(lockDir, 0700); err != nil {
```
DisposaBoy: This mkdir is outside the loop below, so if another process unlocks after this, they might also remove the directory and we will fail to create the lock file.

I reworked the lock acquisition in https://github.com/DisposaBoy/certmagic/commit/b5e59fc352bdf4e6318452d6a315f92ef8827cb3 so I could continue testing.

storage.go
```diff
@@ -100,11 +102,6 @@ type Locker interface {
 	UnlockAllObtained()
```
DisposaBoy: This doesn't appear to be called from anywhere and I don't think it's needed anyway so can be removed.

mholt: This is called by Caddy during shutdown (any signal is not a forceful shutdown).

filestorage.go
mholt: Dang, you're right. Any chance we can still create `FileStorage` without them having to be a pointer? I think the Go compiler takes care of passing in the reference in that case: https://tour.golang.org/methods/4 -- just one less thing to worry about then.

DisposaBoy: That won't work: The method set of `FileStorage{}` only has methods with a `FileStorage` receiver so it doesn't implement the interface (`&FileStorage{}` has both `*FileStorage` and `FileStorage` methods).

storage.go
DisposaBoy: This method doesn't work for cleanup at program start, so it's used only during shutdown and giving this some more thought, I think it can lead to double-unlocking and other bugs:

* Unless it's an immediate `os.Exit()` (an even then), I don't think it's safe to unlock any of the keys: If we unlock the key, the code that locked it will continue running and will eventually call unlock again.
* The existing running code might also go on to change some state, but another process can now acquire the lock and proceed to change the same state.
* Given the above points, it's unsafe to unlock keys locked by our own process (outside the code that locked it).
* The only case (I can think of) in which our process's code could fail to unlock the key is if it panics, and we have `defer` for that.
* The simplest solution I can think of would be to just wait for all code to come back to a safe point (after unlock) - maybe with some context telling it we're shutting down. At that point there shouldn't be any case in which `UnlockAllObtained()` has anything to unlock.

After reading the code some more I saw a couple more bugs. I will comment on those separately.

DisposaBoy:
> @DisposaBoy Would you like collaborator privileges so you can push any fixes to my branch and be credited for it in the commit history? Or I can just make the corrections and update this PR. Let me know!

I don't mind; whichever option's easier for you.

filestorage.go
```diff
@@ -284,25 +304,36 @@ type fileStorageWaiter struct {
...
 
 var _ Storage = FileStorage{}
-var _ Waiter = &fileStorageWaiter{}
 
 // staleLockDuration is the length of time
 // before considering a lock to be stale.
 const staleLockDuration = 2 * time.Hour
+
+// fileStorageNameLocksMu guard access to all
+// FileStorage.fileStorageNameLock values.
```
DisposaBoy: Typo: fileStorageNameLock**s**.

filestorage.go
```diff
@@ -284,25 +304,36 @@ type fileStorageWaiter struct {
...
+var fileStorageNameLocksMu sync.Mutex
```
DisposaBoy: This being global means it will block *all* instances, which is probably not a desirable behavior.

filestorage.go
```diff
@@ -199,27 +214,29 @@ createLock:
...
+	return nil
 }
 
 // Unlock releases the lock for name.
 func (fs FileStorage) Unlock(key string) error {
 	fileStorageNameLocksMu.Lock()
 	defer fileStorageNameLocksMu.Unlock()
 
-	fw, ok := fileStorageNameLocks[key]
+	if fs.fileStorageNameLocks == nil {
```
DisposaBoy: The map is never *written to* so it doesn't need to be initialized.

filestorage.go
```diff
@@ -239,15 +256,18 @@ func (fs FileStorage) Unlock(key string) error {
...
 	return nil
 }
 
-// UnlockAllObtained removes all locks obtained by
-// this instance of fs.
+// UnlockAllObtained removes all locks obtained
+// by this instance of fs.
 func (fs FileStorage) UnlockAllObtained() {
-	for key, fw := range fileStorageNameLocks {
+	if fs.fileStorageNameLocks == nil {
```
DisposaBoy: `fs.fileStorageNameLocks` is accessed without holding `fileStorageNameLocksMu`

filestorage.go
```diff
@@ -199,27 +214,29 @@ createLock:
...
-	fw.wg.Add(1)
-	fileStorageNameLocks[key] = fw
-
-	return nil, nil
+	// cool, we got the lock right away
+	return nil
 }
 
 // Unlock releases the lock for name.
 func (fs FileStorage) Unlock(key string) error {
```
DisposaBoy: typing again after closing the tab....

I think the lockDir removal might be unsafe.
I don't know about on Windows, but on `*nix`, `os.Remove()` should fail if the directory isn't empty, so you don't need to open and read its contents - which is racy because another process might come along and change things between us reading the filenames and calling `os.Remove()`.

In either case, I think the whole parent directory removal is possibly racy:

* we remove the file
* other process proceeds since the file no longer exists
* other process calls mkdirall
* we proceed to remove the directory
* other process fails to create file in the now non-existent directory

I'm not sure you need to remove it, though, because it's always the same directory and it's just going to get re-created anyway.

filestorage.go
mholt: Alright. That's something I think I can live with. (using &FileStorage)

filestorage.go
mholt: fileStorageNameLocksMu protects the _elements_ of the map, not the map itself.

filestorage.go
mholt: Good point!

filestorage.go
mholt: Yes, but the locks are brief, and this was done to avoid copying mutex values also.

mholt:
At this point, I think it would be easier for both of us if I just make you a collaborator -- go ahead and push commits to my branch and then we can fix this together.

storage.go
mholt: Interesting thoughts -- I'd be happy to review a potential solution if you want to submit one!

filestorage.go
DisposaBoy: It still needs synchronization because it's being written to, e.g. inside `Lock()` https://github.com/mholt/certmagic/blob/b1ccc785a5fd3bcb1f9046a19ea2f11cca0770bc/filestorage.go#L163

filestorage.go
DisposaBoy: There shouldn't be any possibility of copying the mutex (and `get vet` should catch that).

filestorage.go
```diff
@@ -192,39 +186,41 @@ func (fs *FileStorage) Lock(key string) error {
 	fs.fileStorageNameLocks[key] = fw
 	fileStorageNameLocksMu.Unlock()
 
-	var checkedStaleLock bool // sentinel value to avoid infinite goto-ing
+	for {
+		// parent dir must exist
+		if err := os.MkdirAll(lockDir, 0700); err != nil {
+			fileStorageNameLocksMu.Unlock()
```
mholt: I believe this Unlock will panic, since it was unlocked on line 187/193 above.

DisposaBoy: oops.. fixed

filestorage.go
```diff
@@ -149,125 +152,124 @@ func dataDir() string {
...
-			// check to see if the lock is stale, if we haven't already
-			if !checkedStaleLock {
-				checkedStaleLock = true
-				if fs.lockFileStale(fw.filename) {
-					log.Printf("[INFO][%s] Lock for '%s' is stale; removing then retrying: %s",
-						fs, key, fw.filename)
-					os.Remove(fw.filename)
-					goto createLock
-				}
+			info, err := os.Stat(fw.filename)
```
mholt: I noticed that this now leaves the `lockFileStale()` method unused -- was this intentional? I am ready to push a small commit to restore this if that is desirable.

filestorage.go
```diff
@@ -149,125 +152,124 @@ func dataDir() string {
...
+			switch {
+			case err != nil:
```
mholt: I think this empty case will cause the loop to spin, perhaps if there is an error Stat'ing we should just assume the lock is stale, since it's apparently useless if we can't read it?

filestorage.go
DisposaBoy: I didn't use `lockFileStale()` because it returns `true` if stat fails (which isn't the same as the file existing being stale).

mholt: Well, that function returning `true` has the same net result as this change: the loop will simply continue.

I'll push a commit in a moment that adjusts based on this feedback, plus a little of my own yak shaving :smile:

mholt: p2 cannot create its own file if p1 already created it: the file creation is atomic (assuming the file system is correctly implemented) because of the O_EXCL flag. Thus p2 will not create its own file and instead wait for the lock to be released.

filestorage.go
```diff
@@ -149,125 +152,124 @@ func dataDir() string {
...
+			default:
+				time.Sleep(1 * time.Second)
 			}
+		default:
+			// otherwise, this was some unexpected error
 
-			// if lock is not stale, wait upon it
-			return fw, nil
+			// we called wg.Add(1) above but didn't actually acquire the lock
+			fw.wg.Done()
```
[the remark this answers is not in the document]
DisposaBoy: It does indeed... the return after mkdir also has the same problem.

filestorage.go
mholt: I don't think that will work either unless the Chtimes call is atomic with the creation of the file...

mholt: I know this is less than ideal, but at least consider this:

Since certificate operations are very infrequent (currently once about every 60 days, only lasting a few seconds), and the consequences of a highly improbable (admittedly not impossible) race are essentially nil (in the event of a concurrent issuance or renewal, Let's Encrypt will still allow both -- it's just if it happens often for a higher number of subdomains that they throttle), maybe the locking doesn't have to be perfect, honestly? And even if the rate limit _is_ hit (which I already think is very unlikely), it would be even _less_ likely the next time, and would still result in no downtime since renewals can be tried for the next 3-4 weeks.

Just something to consider.

But of course I'm down to make this as perfect as possible (within practicality).

filestorage.go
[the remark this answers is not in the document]
mholt: I just noticed that too. I have a nearly-ready commit here that should address this issue by calling Unlock before `Wait()` or `return` in the various code paths, keeping the creation of the fileWaiter atomic while also cleaning it up if obtaining the lock fails.

But I know you said you had some ideas for Unlock, so I'll wait before committing.

filestorage.go
DisposaBoy: I *think* `Unlock` suffers from the same `os.Remove` race we've seen elsewhere in the face of a stale lock - as unlikely as it is:

* p2 sees the lock is stale
* p2 removes it
* p2 creates its own lock
* p1 completes
* p1 removes the file
* p3 is free to proceed

I think a better strategy for the plain `os.Remove` is:

* create another file e.g. `key.unlock` first
* if it succeeds, we take ownership of unlocking:
*  * remove key.lock
*  * remove key.unlock
* otherwise, another process is doing the unlocking

There is a very small risk that the app crashes between `remove key.lock` and `key.unlock` so `Lock` would need to remove `key.unlock` when it successfully creates `key.lock`.

And now you know why I'm not a fan of `lock files`.

DisposaBoy: I agree the locking doesn't need to be perfect (given the use-case), so don't let me be a blocker :). Considering I appear to be the first one to bring up the locking issues, it probably means it's not a big problem.

mholt: I think I understand. I'm willing to accept the infinitesimally small risk_ that the program will crash in that critical section, because even then the consequences won't even be tangible, I think.

Do you want to try implementing those changes? If so, I will hold off! (In fact I have to go do my thesis work for the day, so I won't get around to it for a bit anyway.)

DisposaBoy: I'll add it to my list for tomorrow.

filestorage.go
```diff
@@ -149,125 +152,112 @@ func dataDir() string {
...
+		}
+		// we'll just assume the lockfile exists
+		info, err := os.Stat(fw.filename)
+		switch {
+		case err != nil:
+			// we assume the lockfile no longer exists; if we're now able to create it, great!
+			err := createLockfile(fw.filename)
+			if err != nil {
+				// we called wg.Add(1) above but didn't actually acquire the lock
+				fw.wg.Done()
```
mholt: This still leaves the waiter "dangling" in the map (it should probably be deleted from the fileStorageLocks map, no? - that might require extending the lock on its mutex down to this function's final return _or_ `Wait`...)

filestorage.go
```diff
@@ -149,125 +152,112 @@ func dataDir() string {
...
-
-			// if lock is not stale, wait upon it
-			return fw, nil
+			return err
+		case fileLockIsStale(info):
+			log.Printf("[INFO][%s] Lock for '%s' is stale; removing then retrying: %s",
+				fs, key, fw.filename)
+			removeLockfile(fw.filename)
+		default:
+			time.Sleep(1 * time.Second)
```
mholt: Is it intentional that we're basically repeating the logic of `Wait()` here?

filestorage.go
```diff
@@ -149,125 +152,112 @@ func dataDir() string {
...
+			// we assume the lockfile no longer exists; if we're now able to create it, great!
+			err := createLockfile(fw.filename)
+			if err != nil {
+				// we called wg.Add(1) above but didn't actually acquire the lock
+				fw.wg.Done()
 			}
-
-			// if lock is not stale, wait upon it
-			return fw, nil
+			return err
```
mholt: One thing I just realized about this change here is that it could erroneously return an error.

For example, to get here we first tried to create the lock on disk which failed, presumably because it already existed; then we tried to stat it, which failed, so it was probably removed; then we tried to create it again, which failed, so... what does that mean? Maybe it was created again by another process, but here we just return an error, even though the correct thing to do would be to wait longer.

I'll push a commit soon which addresses this, I think. Hopefully for our final round of reviews! :+1:

mholt:
Okay, I've pushed a commit that _I think_ addresses all outstanding issues mentioned in our reviews up to this point.

I decided to get rid of the existing notion of the `fileStorageWaiter` (it was confusing and buggy I think) and replace it with a `fileLockWaiter` which is just a channel that is used to block until it is closed. Even this is not strictly necessary, but if there are many threads in the same process that are vying for this lock, this waiter reduces file system hits.

If that optimization is not valuable (i.e. if we can accept many threads polling disk to check the status of the lock), I think we can do away with the `fileLockWaiter` entirely (and thus most of the logic in the `default` case of our switch in the loop to create a lock). Would love to know what you think!

storage.go
mholt: I eliminated this method.

DisposaBoy:
Unfortunately I didn't have much time today so I didn't do a very thorough review but the new code is very similar to what I had in mind.

I also wondered if we needed the internal lock (`lockWaiters`) at all. It seemed to just add more complexity with possibly no big performance gain.

mholt:
Yeah, like I was saying, it's just to reduce extra polling. If the occasional multiple polling is OK, then we can nuke it.

filestorage.go
```diff
@@ -118,191 +120,221 @@ func (fs FileStorage) Stat(key string) (KeyInfo, error) {
...
+				fw.wait()
+				continue
+			}
+			waitChan := make(chan struct{})
+			fs.lockWaiters[key] = waitChan
+			fs.lockWaitersMu.Unlock()
+
+			// poll until the lock is available; ours should be
+			// the only thread in this process polling the disk
+			for time.Since(start) < staleLockDuration {
```
DisposaBoy: I think there's a small possibility that this condition can be false before we enter the loop and we'd fail to close `waitChan`.

DisposaBoy:
The test app runs fine and nothing else stands out, so apart from my last comment it LGTM.

DisposaBoy:
> Yeah, like I was saying, it's just to reduce extra polling. If the occasional multiple polling is OK, then we can nuke it.

I'd personally nuke it. It's always easy to bring it back later if there's a performance issue.

mholt:
Sounds good -- I'll do that right now, stay tuned (just a few minutes) for the final commit. Then when you change your review to Approved I can merge this!

mholt:
You're right, this is much better! Thanks for the iterative reviews. This PR also removes almost 30 lines of code! Ready to merge.

PR #13, opened 2019-01-07
exported KeyBuilder.Safe to use with external pkgs

securityclippy:
Really simple change here.  Simply changing the `KeyBuilder.Safe()` function to be exported instead of internal to certmagic.  This allows other storage implementations to take advantage of the method without re-implementing the entire type.

Hope that's OK and makes sense!


securityclippy:
Wahoo, thanks @mholt !  Used it in the S3 implementation which I will link up shortly.  Much appreciated!

PR #23, opened 2019-01-29
Support for non SNI clients

### LINKED ISSUE #22: Non SNI Certificates
Commit c1d472b46046ee329c099086d689ada0c44d56b0 made certmagic set the listeners IP address as the certificate name in case the server name has not been provided (not using SNI).

In a current case I have an app with certmagic running in docker. If a non SNI client is connecting, certmagic sets the private `172.17.X` ip address of the docker container as the certificate name and since there is no certificate managed under that name it logs the error "no certificate available for 172.17.X".

Can we add a switch to the configuration to set a fallback certificate name for such cases? In my case I run certmagic with a single domain and would like to use the same certificate also for non SNI clients.

Maybe something like:

```go
certmagic.NonSNIDomain = "example.com"
```

In case the option is not set, we can fallback to the current behaviour.

256dpi:
Fixes #22 .


mholt:
Pinging @ghoeffner @orenyomtov @Zenexer for feedback since we're actually making some headway on this issue over here.

ghoeffner:
> Pinging @ghoeffner @orenyomtov @Zenexer for feedback since we're actually making some headway on this issue over here.

To my understanding this change would mean that support for non-sni is restored and it can be enabled via the config? (I haven't had a chance to look at the full PR yet)

mholt:
@ghoeffner 

> To my understanding this change would mean that support for non-sni is restored and it can be enabled via the config? (I haven't had a chance to look at the full PR yet)

Yes, this in conjunction with changes in Caddy that I am currently working on locally, which will not be part of this PR. This requires the site owner whitelist a specific TLS config / hostname to be used as a default, otherwise there will not be one.

But I need to figure out how to best handle these situations, hence my request for feedback.

(We need to figure out whether we differentiate between empty SNI and non-empty, non-matching SNI. Right now I guess I assume that a default host will only apply when SNI is missing.)

mholt:
@256dpi Please try the latest commit which I just pushed, and verify that it works for you. If so, I am ready to merge it.

mholt:
Okay actually I'm just gonna merge this so I can keep working on Caddy, and if this is a bad change we can fix it later. But I think/hope it's a good one.

256dpi:
@mholt Sorry was busy. But the PR looks good. Thanks!

PR #24, opened 2019-01-30
Allow disabling certificate renewal

### LINKED ISSUE #21: Disable certificate renewal
Ability to disable certificate renewal.

I'm using the dns challenge since I need wildcard certificates, the problem is that every server needs to know the credentials for my dns provider, which I find unnecessary since I could have a single server do the renewal, or something hosted on some serverless platform, so because of this it would be awesome to have the ability to disable renewal on the rest of the servers.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
I'm setting up a dns provider but passing in invalid credentials, which will probably become a problem when certmagic will decide it's time to renew the certificates.

paolobarbolini:
Setting `Config.RenewalDurationBefore` to a negative value will
disable certificate renewal.

Closes #21

### DIFF
--- a/certificates.go
+++ b/certificates.go
@@ -68,6 +68,9 @@ func (c Certificate) NeedsRenewal() bool {
 	if len(c.configs) > 0 && c.configs[0].RenewDurationBefore > 0 {
 		renewDurationBefore = c.configs[0].RenewDurationBefore
 	}
+	if renewDurationBefore < 0 {
+		return false
+	}
 	return time.Until(c.NotAfter) < renewDurationBefore
 }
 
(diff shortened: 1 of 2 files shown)


mholt:
Now that I think about it, changing `RenewDurationBefore` is not the right solution. That parameter configures when Caddy should consider "renewing" the certificate, which it always does first by trying to just load it from storage, and if it's already renewed in storage then it doesn't need to initiate a new renewal.

But it still needs to load it from storage.

I think we need a separate parameter to disable initiating new certificate renewals.

PR #26, opened 2019-03-01
added certmagic.HTTPSWithSelfSigned() 

### LINKED ISSUE #25: add selfSignedCertificate
it would be usefull to have the "newSelfSignedCertificate" code migrated from caddy to certmagic.

In order to test applications in local docker env's.
requesting, creating, signing and managing a selfsigned for an application is still pain and needs some more magic!

I think a good solution would be to have something simular like certmagic.HTTPS() which creates and uses a selfsigned

mschneider82:
fixes #25 

HTTPSWithSelfSigned() is the same like HTTPS() but creates and uses selfsigned certificates instead of letsencrypt.

migrated selfsigned.go from caddy

### DIFF
--- a/certmagic.go
+++ b/certmagic.go
@@ -134,6 +134,94 @@ func HTTPS(domainNames []string, mux http.Handler) error {
 	return httpsServer.Serve(hsln)
 }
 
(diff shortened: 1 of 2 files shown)


mschneider82:
closed because #27 seems a better solution

PR #27, opened 2019-03-01
certmagic.SelfSignedCA as certmagic.CA value

mschneider82:
introduced: certmagic.SelfSignedCA as possible certmagic.CA value to create a self-signed certificate with HTTPS()

implemented a prototype for idea in #25

### DIFF
--- a/certmagic.go
+++ b/certmagic.go
@@ -68,10 +68,38 @@ func HTTPS(domainNames []string, mux http.Handler) error {
 	if mux == nil {
 		mux = http.DefaultServeMux
 	}
+	var selfsigned bool
 
-	cfg, err := manageWithDefaultConfig(domainNames, false)
+	cfg := NewDefault()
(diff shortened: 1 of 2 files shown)


mrg0lden:
@mholt will this PR get some love?

mholt:
Perhaps. At a glance, this approach has at least one significant downside: self-signed certificates can't actually be managed; i.e. there is no way to renew them. 10-year lifetime is not exactly a good idea in many situations, and a better approach would be to create a single CA from which all other certs are self-signed.

I'm currently refactoring how CertMagic manages certificates, which has strong implications about how self-signed certificates are handled. The `Manager` interface recently introduced on master should make it possible to implement a SelfSigned certificate manager. But whether that ends up in this repo is still an open question. If the implementation is good enough, then perhaps!

Also it will need tests.

mholt:
Closing, because I think this approach is outdated now...

PR #37, opened 2019-04-01
Change default KeyType to EC256 (P256)

My1:
will also allow IE11 to use GCM ciphers (indirect fix for mholt/caddy#2496 )

### DIFF
--- a/certmagic.go
+++ b/certmagic.go
@@ -474,7 +474,7 @@ var (
 
 	// The type of key to use when generating
 	// certificates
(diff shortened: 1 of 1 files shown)


My1:
LGTM? what does that mean?

midzer:
LGTM = Looks good to me

My1:
I see but well that change was dead simple anyway, the only "issue" at hand was just finding out that it wasnt actually in caddy (and making sure you entered the correct name for the curve as Go calls it EC256 and not P256, XD

PR #48, opened 2019-10-19
Add an Config option for custom ChallengeOptions

joohoi:
This PR adds a new Config option to allow the user to define custom `dns01.ChallengeOption` for their `dns01.Provider`. This gives optionally more control for the advanced users of certmagic.

#### Problem description and example use case:
When running DNS servers behind firewalls with address translation, it may prove difficult for the default validation functions of `lego` to successfully return. This is true even in cases where the validation should succeed. 

The reason for this is that some firewalls struggle with packets from LAN towards WAN IP that actually points back to the LAN. This is usually handled by [split-horizon DNS](https://en.wikipedia.org/wiki/Split-horizon_DNS), but in this use case it proves insufficient.

With this option, the user is able to add a custom DNS resolvers, modify the propagation check rules or change the preCheck validation logic to suit their needs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -579,7 +579,11 @@ func (c *acmeClient) nextChallenge(available []challenge.Type) (challenge.Type, 
 		})
 
 	case challenge.DNS01:
-		c.acmeClient.Challenge.SetDNS01Provider(c.config.DNSProvider)
+		if c.config.DNSChallengeOption != nil {
+			c.acmeClient.Challenge.SetDNS01Provider(c.config.DNSProvider, c.config.DNSChallengeOption)
+		} else {
+			c.acmeClient.Challenge.SetDNS01Provider(c.config.DNSProvider)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


joohoi:
#### PoC adapted from the example in README, disabling the preflight check logic
```
import (
    "github.com/go-acme/lego/v3/challenge/dns01"
    "github.com/go-acme/lego/v3/providers/dns/dnsimple"
)

provider, err := dnsimple.NewDNSProvider()
if err != nil {
	return err
}

certmagic.Default.DNSProvider = provider
certmagic.Default.DNSChallengeOption = dns01.WrapPreCheck(func(_, _, _ string, _ dns01.PreCheckFunc) (bool, error) {
    return true, nil
})
```

mholt:
Ohhh nice. I actually was thinking that disabling the pre check for all DNS challenges would be a good idea, but let's start here! I'll review this as soon as I get to my computer.

PR #49, opened 2019-10-31
Subject Alternative Names (SAN) Support

aloababa:
Hey,

I'm using your library for a long time now. It's really nice and magic :p

Recently I encountered the need to have Subject Alternative Names in certificate managed by certmagic. So I made a PR to add this feature in certmagic.

I did not want to break the API to have backward compatibility. So I added a new Manager interface in order to keep old code working.


mholt:
Hi, thanks for your PR. But this is a big change, and I don't think this is a good fit for CertMagic, or automated certificate management at all.

It would have been good to open an issue to discuss first.

To clarify, CertMagic _only_ uses SANs already. CommonName is deprecated and should not be used. CertMagic supports CommonName if given a certificate with a CommonName, but does not create certificates with CommonName.

Why do you need multiple names per certificate?

Using multiple names per certificate in an automated environment is discouraged, see: https://github.com/https-dev/docs/blob/master/acme-ops.md#use-one-name-per-certificate

Further, this change is not correct for managing multiple names per certificate. It is not simply a matter of turning a single string into a slice of strings. You also need to redesign and re-implement how certificates are stored, you need to redo the renewal logic, and you need to handle all the complexities mentioned in that linked document (and probably more) that come from mapping multiple domains to a single certificate. This change neglects those details, and will likely cause errors and even downtime.

This has been discussed many times before, by the way -- back when this was still hard-coded into Caddy. It turns out that there is almost never a practical use case or a good justification for using multiple SANs per certificate when the management of the certificates is automated.

I'm going to save you some trouble and close this now, but feel free to continue the discussion if you'd like.

hekmon:
I can see a valid solution where multiples names can be usefull and valid: failover scenarios.

In this case, the certificate will be issued to the domain pointing to the virtual IP. This does not prevent your servers from having their own IP and DNS record pointing to them. In that case, connecting directly to one or another server (that could potential hold the virtual IP, or not) using this record (for debug/monitoring/metrics/you name it) will yield an invalid certificate forcing you to switch to insecure which will deactivate not only name verification but other TLS related checks (ex: ciphers).

By having alternate names, each server can have a certificate for their own DNS record but also for the DNS record pointing to the virtual IP. That way, you can use the VIP or make a direct connection and still keep secure TLS connection. Here another solution could be wildcards... IF both records share the same DNS zone level, which is not always (never ?) the case.

As your document states, it is recommended to use 1:1 when possible. But this is a should, not a must. Lego (which certmagic relies on) perfectly support them: it is a shame certmagic does not even consider to add support for them (documentation could indicate that the 1:1 usage should be preferred when possible).

mholt:
@hekmon 

I don't follow. 

> In this case, the certificate will be issued to the domain pointing to the virtual IP. 

Can you explain what a "virtual IP" is in this case?

> This does not prevent your servers from having their own IP and DNS record pointing to them.

It's not clear what "This" refers to.

> using this record (for debug/monitoring/metrics/you name it) will yield an invalid certificate forcing you to switch to insecure which will deactivate not only name verification but other TLS related checks (ex: ciphers).

This is where I really start to get lost. We're talking about the provisioning/management of certificates here, not the serving of them to clients, right?

> That way, you can use the VIP or make a direct connection and still keep secure TLS connection.

So, are you talking about a proxy?

> As your document states, it is recommended to use 1:1 when possible. But this is a should, not a must. Lego (which certmagic relies on) perfectly support them: it is a shame certmagic does not even consider to add support for them (documentation could indicate that the 1:1 usage should be preferred when possible).

The document states the reasons why M:1 is discouraged:

> Using M:1 has little benefit in automated settings, complicates the addition and removal of names, slows down certificate issuance, and increases the chance of validation errors. There will be greater logical and operational complexity by combining many names onto one certificate. In addition, all those names necessarily share a single private key, so if one key is lost, all those sites need a new certificate with a different key. Multi-SAN certificates also have a larger size, so they slow down TLS handshakes.

Lego supports M:1 because it is an ACME client, not a certificate _manager_, which is what makes CertMagic unique: it actually manages certificates over a long period of time. Lego is a run-and-done command.

hekmon:
This is not quite a proxy. Given the following example:

Server A - IP1 - DNS X (direct access)
Server B - IP2 - DNS Y (direct access, backup server)
Virtual IP - IP3 - DNS Z (High availability access, this IP can move between server A & server B using, for example, heartbeat)

Regular access to the service normally use DNS Z which points to the virtual IP 3 for High Availability. The answering server could be A or B so both need a certificate for the DNS Z. But by having only one name in it, direct connection to A or B (debug/monitoring/...) using DNS X or Y will yield a certificate valid only for Z making the connection invalid and forcing the client to go insecure.

By having Server A request a certificate for both DNS records X & Z and Server B for both DNS Y & Z regular/HA access using DNS Z AND direct access using DNS Y or Z will always be valid as they should be.

In that case multiples names are not aliases (which is mainly used for SAN: example.tld/www.example.tld) but instead two valid & differents entry points that servers A & B can rightfully hold.

I totally get that multiples names do not fit your vision and so are not implemented in certmagic. I just wanted to point out that there are valid cases that need multiples names.

mholt:
Hmm, I'm still not sure I follow, though: the DNS name (domain name) in a certificate is irrelevant of the underlying IP address or the domain's DNS records.

I think if I am going to understand your use case, a concrete example would help. It would need to clearly demonstrate:

1. What the actual domain names and DNS records are
2. Which machines are serving which domain names
3. For machines that serve more than one domain name, why they can't use more than 1 certificate
4. How all of this relates to a certificate's validity (acceptance by the client)

You seem to be suggesting that without this feature, it is impossible for your sites to be served without a proper certificate. But the point of certificates is that they don't care what your architecture is, as long as you serve a non-expired cert with the correct name.

hekmon:
IP was only relevant because of the VIP: DNS Z always point to IP3 but IP3 can be either on Server A or B so DNS Z can either reach Server A or Server B depending on the current system state.

I'm not sure I can make an example more straight forward but I'm not saying this is the only solution/impossible otherwise: multiples certificates could be a workaround but it would be that for me: a workaround.

I would use multiples certificates in a multi tenant scenario (for example a web server holding completely different/distinct web sites). But in case of single tenant service accessible by multiples DNS records, I would go for one certificate & multiples SAN. I understand that this is a personal preference.

But here what you asked I hope it will help make my previous example clearer:

1. `service.organization.tld` `srvA.internal.organization.tld` `srvB.internal.organization.tld`
2. Server A can serve `service.organization.tld` and `srvA.internal.organization.tld` for the exact same ressources underneath. Server B can serve `service.organization.tld` and `srvB.internal.organization.tld` for the exact same resources underneath too.
3. They can. This is just "ugly" :) (single tenant / multi tenant)
4. Not sure to get that, sry. Let's try anyway: I just never want the client (external or internal) to go insecure. This need one certificate with multiples SAN or multiples certificate with a single SAN (which technically works but... cf 3.).

Again, I'm fine with the "not my vision" answer, this is your project and I too kindly refused PR that was not aligned with what I wanted to achieve. I just don't like to use workarounds when a cleaner solution exists :]

mholt:
Okay, I think I understand a little better now. Basically you advocate for M:1 certificates because you view 1:1 as "ugly", not because 1:1 is not an option.

I guess I can understand that, but with CertMagic you never have to see, edit, or even think about the certificates, so it doesn't matter how many certificates are under the hood. The functionality is the same, and the benefits are much greater with 1:1 certs.

Trust me, compared to the logic of managing M:1 certificates correctly and reliably, 1:1 is a much cleaner solution, I guarantee it.

mholt:
By the way, how did you all find this PR so fast?  It seems like @cedriclaudrel, @dekonnection, @jonathanmarsaud, @jaypif, and yourself, @hekmon, all visited it within about 10 minutes of its opening. I'm curious if this repo is attracting any attention, and where the traffic is coming from suddenly. It's usually much quieter around here.

hekmon:
You could find a clue under the "Organizations" on our profils ;D

We do a lot of automatization, micro services & custom tools in Go. We poked around our own acme client some time ago (not fan of certbot) but as a small team we quickly moved to lib like lego. Several of us were keeping an eye on Caddy back then: this is how certmagic came under our radar. We decided to give it a go as it offered a lot of wrapping work on top of lego.

mholt:
Cool, thanks for letting me know. Always nice to know how people find and use a project!

PR #57, opened 2020-01-17
Use HTTPSPort for redirecting in case a custom port is defined

Freeaqingme:


mholt:
Can you be more specific? What problem are you having?

Freeaqingme:
Sorry, I definitely can. On my local dev environment, I do:
```
	certmagic.HTTPPort = 2480
	certmagic.HTTPSPort = 2443
```

But if I connect to http://localhost:2480, I am redirected to https://localhost. The https listener does actually listen on port 2443, so I'm redirected to a port where nothing is listening. This PR will ensure that the user is redirected to the port where the SSL listener is actually listening on.

mholt:
I see... thanks for explaining.

So this is actually tricky. The HTTPS port is a defined standard: 443. The purpose of the HTTPSPort variable is if you have port forwarding set up from port 443, for internal use, for example solving the ACME TLS-ALPN challenge.

Thus, the variable is intended for internal ports, not external ones, since your clients won't also have HTTPS configured on port 2443.

Freeaqingme:
Yeah, I realize that it risks breaking BC for users who rely on this specific behavior (though I don't think it's been documented as such). For me I only use a different port on dev and then it should be relatively safe.

If you believe we should just close this PR, feel free to go ahead. In case you feel changing behavior, we could introduce another param to specify the port you'd want to redirect the user to. But if/how you want to approach that I'll happily leave up to you as well :)

mholt:
Okay. Yeah, I think I'll close this, to keep the variable to its intended purpose. If you need this kind of redirect though, it's not hard to implement one manually. :)

PR #60, opened 2020-02-10
Support custom tls NextProtos

lvht:
The h2 has the problem known as “head-of-line blocking”. 

It is not a good idea to enable h2 in a network with high package lose rate.

### DIFF
--- a/config.go
+++ b/config.go
@@ -79,6 +79,11 @@ type Config struct {
 	// challenge to succeed
 	AltTLSALPNPort int
 
+	// NextProtos is a list of supported application level protocols, in
+	// order of preference.
+	// The default NextProtos is ["h2", "http/1.1", "acme-tls/1"].
+	NextProtos []string
+
 	// The DNS provider to use when solving the
 	// ACME DNS challenge
 	DNSProvider challenge.Provider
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the PR.

But, I don't think I want to expose the TLS configuration in the certmagic.Config struct. The Config struct is concerned primarily with ACME operations, and the TLSConfig method is _supposed_ to be opinionated (as documented). The caller could simply change the returned TLS config instead. (That is the point.)

Also, this is not really accurate, or the whole picture:

> The h2 has the problem known as “head-of-line blocking”.

HTTP/2 actually does solve this at the application layer and does not suffer from HoL blocking. However, there's still TCP HoL blocking, which neither HTTP/2 nor HTTP/1.1 solve.

lvht:
@mholt Thanks for reply.

I did not read the readme carefully. And I should use the following code
```go
// use this to configure a TLS listener
tlsConfig := magic.TLSConfig()
tlsConfig.NextProtos = []string{"http/1.1", tlsalpn01.ACMETLS1Protocol}
ln, err := tls.Listen("tcp", ":443", tlsConfig)
```

BTW, if the network package lose rate is high, having multiple http/1.1 session is better than on h2 session with multiple stream, especially for occasion of https proxy.

Thank you.

mholt:
> BTW, if the network package lose rate is high, having multiple http/1.1 session is better than on h2 session with multiple stream, especially for occasion of https proxy.

Depends, but that is due to multiple TCP streams -- which both HTTP/1 and HTTP/2 still rely on -- which are less efficient and slower due to more round trips. It _might_ be better in some edge cases, but it's very much a narrow/specific optimization these days.

lvht:
> Depends, but that is due to multiple TCP streams
Yes. Multiple TCP streams beats one, and this is why HTTP/3 developing.

But is the client is standard browser and the ALPN offers h2, the browser will always use h2, which make multiple TCP streams impossible.

Thanks.

PR #73, opened 2020-05-13
chore: update go version declared in go.mod

x1nchen:
Function handshake.DefaultCertificateSelector uses the field
`*tls.ClientHelloInfo.SupportsCertificate` which only supported in go 1.14

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/caddyserver/certmagic
 
-go 1.13
+go 1.14
 
 require (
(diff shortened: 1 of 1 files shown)


x1nchen:
ref: crypto/tls: implement (*ClientHelloInfo).SupportsCertificate · golang/go@dd01738: https://github.com/golang/go/commit/dd0173845f939dc532e8e8348b8360914dbca216

x1nchen:
emmm, I suppose we'd better add some expaination to document why this module is only supported by go 1.14.

PR #74, opened 2020-05-25
Add Shutdown that accompanies HTTPS

vmihailenco:


mholt:
If you need Shutdown, just make your own HTTP server and call Listen() instead -- much simpler, no code changes required, and gives you that degree of control. Thanks!

vmihailenco:
That is more like - every user is supposed to copy the whole HTTPS method and maintain it over time - how that is "much simpler" is a mystery.

mholt:
No, `HTTPS()` is meant for simple applications that don't need a bunch of maintenance -- believe it or not, most custom Go programs don't stop their servers mid-process, and those that do use a http.Server. `HTTPS()` is meant to be very opinionated, since it is the highest-level function geared toward the majority of users that simply need to run an HTTPS server. If you need more advanced capabilities, use the lower-level interfaces.

Thank you for the contribution -- in the future, it is always a good idea to open an issue to propose a change before a pull request so it can be discussed first. :) (This is true for any project.)

Issue #76, opened 2020-06-03
Wildcard certificate is not loaded from storage


MichaelMure:
## What version of the package are you using?
github.com/caddyserver/certmagic v0.11.0

## What are you trying to do?

I'm attempting to setup a cluster configuration where go process do the TLS termination and share a common storage. However (and I suspect the problem start here), the certificates creation is done in another process connected to the same storage, by calling `config.ManageSync()` and exiting.

(On a side note, it'd be nice if a lower level function would be exposed to check/generate/renew certificates without the higher level stuff that `ManageSync` imply).

## What steps did you take?

After making sure that a wildcard certificate exist for `*.example.com`, I did a HTTPS request to `foo.example.com` on one of the nodes connected to the common storage.

Following what happen with a debugger, it seems to me that the in-memory cache is checked first. This fail as the node doesn't have any certificates at that point. 

As a fallback, the storage is then checked. However, **only the exact match for `foo.example.com` is attempted**.

(Side note again, an `OnDemand` config needs to be created for that to work. However, I do not want to generate new certificates, only load them from the disk.)

## What did you expect to happen, and what actually happened instead?

I expected my wildcard certificates to be loaded.

## How do you think this should be fixed?

I would think the storage should be checked for both `foo.example.com` and `*.example.com`. Possibly in parallel.

## Bonus: What do you use CertMagic for, and do you find it useful?

Because it's the only way I could find to solve my problem. That's no small feat.

MichaelMure:
Note: it seems like the in-memory cache is checked with many level of wildcard (`*.example.com` then `*.*.com`). As far as I know this is not necessary as TLS certificates can only have a single left-most level of wildcard.

Same thing for the storage, only one possible wildcard needs to be checked as far as I know.

mholt:
Hm, okay, yeah, so I think this only makes sense for On-Demand.

> As a fallback, the storage is then checked. However, only the exact match for foo.example.com is attempted.

I suppose we could make this an optional behavior, a boolean field in the OnDemandConfig struct; if enabled, try loading wildcard permutation of the domain name. We probably shouldn't assume that a wildcard should always be used if that's what is available -- usually, on-demand TLS and wildcards are not used together, so I'm curious about your need for this.

[...]

PR #77, opened 2020-06-05
Add a way to configure cert loading from the storage

MichaelMure:
This is a proposal PR to address https://github.com/caddyserver/certmagic/issues/76

Note: I don't fully understand the assumption made in CertMagic so this might be incorrect/problematic in some places. I also didn't fully test it. This PR is to show what changes I suggest.

In term of DX, changes are as follow:
- the OnDemand concept is restricted to generating certs from the CA instead of this + loading from storage
- an extra configuration struct is added to control how storage certs are loaded
- exact match certificates are still loaded as previously
- loading multi-level wildcard certificates is gated by an extra config flag to avoid the performance penalty, even if you want to load wildcard certificates

### DIFF
--- a/certmagic.go
+++ b/certmagic.go
@@ -247,7 +247,7 @@ func ManageAsync(ctx context.Context, domainNames []string) error {
 // containing the names passed into those functions if
 // no DecisionFunc is set. This ensures some degree of
 // control by default to avoid certificate operations for
-// aribtrary domain names. To override this whitelist,
+// arbitrary domain names. To override this whitelist,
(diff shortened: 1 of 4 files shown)


MichaelMure:
@mholt Is that something you would consider?

mholt:
I'll take a look at it as soon as I get a chance, thanks!

MichaelMure:
No pressure :)

mholt:
I forget when exactly it happened, but loading wildcards from storage is already implemented. Closing this.

PR #90, opened 2020-08-23
cache: remove unused method getFirstMatchingCert

jimen0:
### What does this PR do?

Removes unused `(*Cache).getFirstMatchingCert()` method.

### What does it improve?

Removes dead code.

### How was it tested?

`go test -race -v ./...`

### What Go version did you use when testing it?

`1.15` and `tip`.


mholt (approved):
Thanks for finding that! Guess I left it dangling when I refactored some stuff.

PR #96, opened 2020-09-12
dnsutil: Don't error when checking record with preexisting values

delthas:
When checking whether a new DNS TXT record is deployed, as part of the
DNS challenge procedure, checkAuthoritativeNss is called in a loop until
the requested TXT value is found in one of the records, or until a
timeout.

Previously, if there were other DNS TXT records for the same FQDN, the
call to checkAuthoritativeNss failed and the whole DNS challenge was
canceled. This means for example that if there was any previous
_acme-challenge TXT for the domain, the DNS challenge would always fail.

This fixes this issue by not returning an error, but instead returning
not ready, when there are other values returned by that DNS TXT record
request.


delthas:
I have tested this successfully on my server by running certmagic with a DNS solver with a preexisting _acme-challenge TXT record. It didn't work before and worked after the patch 🙂

mholt (changes requested):
Thanks, I have also seen that error before but that was before we split from lego so I didn't feel motivated to change anything.

This can get even simpler: we don't need `found`, because `if record == value` we can just `return true, nil` right away. (I'm not sure why it was written this way, like if there was a reason behind it. But using `found` seems unnecessary.)

delthas:
The current behaviour waits for all authoriative nameservers to have the record before continuing with the challenge. (ie If any server does not have the record, return false.) Using return true right without `found` change the behaviour to: wait until any nameserver has the record.

mholt (approved):
Ah true, I forgot about the outer loop.

I keep thinking about this and don't see a reason why this isn't good logic, so let's give it a try. Thanks!

PR #103, opened 2020-10-05
In HTTPChallengeHandler, allow pass nil as h http.Handler to ignored unhandled HTTP requests

wheelcomplex:
When use certmagic with fasthttprouter we will route all "/.well-known/acme-challenge/*token" requires to certmagic HTTPChallengeHandler and want to ignore  invalid challenge requests which may not handled by certmaigc.HandleHTTPChallenge.

Actually, it is for https://github.com/wheelcomplex/magicgate :XD

Many thanks.


go.sum
```diff
@@ -79,4 +79,4 @@ gopkg.in/errgo.v2 v2.1.0/go.mod h1:hNsd1EY+bozCKY1Ytp96fpM3vjJbqLJn88ws8XvfDNI=
 gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 honnef.co/go/tools v0.0.1-2019.2.3 h1:3JgtbtFHMiCmsznwGVTUWbgGov+pVqnlf1dEJTNAXeM=
-honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt0JzvZhAg=
+honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt0JzvZhAg=
```
francislavoie, line 82: Nothing actually changed here, could you remove `go.sum` and `go.mod` from your PR? I wonder if line endings were changed 🤔

httphandler.go
```diff
@@ -28,13 +28,16 @@ import (
 // cache backed by a functional storage facility, since that is where
 // the challenge state is stored between initiation and solution.
 //
-// If a request is not an ACME HTTP challenge, h will be invoked.
+// If a request is not an ACME HTTP challenge, h will be invoked (if h is not nil).
+// If h is nil all unhandled HTTP requests will be ignored.
```
francislavoie, line 32:
```suggestion
// If a request is not an ACME HTTP challenge, h will be invoked.
// If h is nil, all unhandled HTTP requests will be ignored.
```

go.sum
wheelcomplex: Thanks for suggestion, I have updated the PR.

httphandler.go
wheelcomplex: done as suggestion :)

mholt:
Can you help me understand why this is useful? Why not just use [`HandleHTTPChallenge()`](https://pkg.go.dev/github.com/caddyserver/certmagic#ACMEManager.HandleHTTPChallenge)? 

> we will route all "/.well-known/acme-challenge/*token"

If you still want to do this, you might want to use https://pkg.go.dev/github.com/caddyserver/certmagic#LooksLikeHTTPChallenge instead -- rather than repeating that logic.

wheelcomplex:
> LooksLikeHTTPChallenge

Thank you for the reply.

For "Why not just use HandleHTTPChallenge()?" , certmagic is using net/http to process http requests, but there is not a HandleFastHTTPChallenge() for the one who use [fasthttp](https://github.com/valyala/fasthttp). 

Currently a [wrap func from fasthttp](https://pkg.go.dev/github.com/valyala/fasthttp@v1.16.0/fasthttpadaptor#NewFastHTTPHandlerFunc) can wrap HandleHTTPChallenge() into a fasthttp handler. Also  [ another wrap func from fasthttp](https://pkg.go.dev/github.com/valyala/fasthttp@v1.16.0/fasthttpadaptor#NewFastHTTPHandler) can wrap HTTPChallengeHandler. And a wrap func always has extra cost.

After review my idea I believe you are right, it's no reason to repeating that logic. A front dispatcher and a wrapped  HandleHTTPChallenge is the best way currently.

I hope I can implement a HandleFastHTTPChallenge for fasthttp later.

Thanks again!

mholt:
Okay, thanks for considering!

What is the actual cost of wrapping a function? I'm willing to bet we're talking nanoseconds, if that, which is microscopic on the scale of network requests.

Issue #92, opened 2020-09-07
Add Config.Unmanage


emersion:
## What would you like to have changed?

I'd like to add a way to stop managing a certificate.

## Why is this feature a useful, necessary, and/or important addition to this project?

Right now there's only a way to add a domain name to the list of managed certificates. This means the number of managed certificates can only grow.

When a certificate is no longer to be used, I'd like to communicate that information to certmagic so that it stops renewing the certificate.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

Workaround: throw away the whole `Config` and re-add all domain names I want certmagic to manage. In the meantime, I just let certmagic manage unused certificates.

emersion:
Hm, I just noticed there is a `context.Context` you can provide to `ManageAsync`. 

> Cancelling ctx will cancel retries and shut down any goroutines spawned by ManageAsync. 

Does that mean cancelling `ctx` will stop renewals and eventually evict the cert from the cache?

mholt:
Good questions!

I admit this is not super clear, especially right now pre-1.0. `ManageAsync()` loads a cert into the in-memory cache -- or if it does not yet exist in storage, it will obtain it -- so that it can be used. That's all it does by itself: the rest is up to the in-memory cache. The cache lifetime is different from that of configs (on which you call `ManageAsync()`) because a config might only live for a brief moment, whereas the cache could be needed for days and weeks and even years. The initial loading or obtaining of the certificate is bound to ctx, but not the whole cache. While a cache is active (`NewCache()`), it maintains the certificates that are in it. This maintenance can be stopped by calling `Stop()`; it is not bound by ctx.

So, to clarify, throwing away configs is fine: they can be ephemeral. But the cache should be singular (for most applications) and long-lived.

Right now there is no exported way to remove a certificate from the cache, only an unexported method:

https://github.com/caddyserver/certmagic/blob/3cc8865d3cd480e50a7f3f0e929a2fa78c55351e/cache.go#L228-L232

We could export it, but what do you want your input to be: just a domain name? It gets complicated because a domain could appear on more than 1 certificate. The cache contains both managed and unmanaged certificates. Do you want to only be able to remove managed certificates, then? Or unmanaged (manually-loaded and manually-renewed) certs too?

emersion:
>The initial loading or obtaining of the certificate is bound to ctx, but not the whole cache.

Okay, that makes it clear that cancelling the context won't be enough.

[...]

PR #106, opened 2020-10-14
Add Config.Unmanage

emersion:
Closes: https://github.com/caddyserver/certmagic/issues/92

Does this match what you had in mind?

Do we also need to do some `Storage` cleanups?

### DIFF
--- a/config.go
+++ b/config.go
@@ -360,6 +360,28 @@ func (cfg *Config) manageOne(ctx context.Context, domainName string, async bool)
 	return nil
 }
 
+// Unmanage causes the certificates for domainNames to stop being managed.
+// If there are certificates for the supplied domain names in the cache, they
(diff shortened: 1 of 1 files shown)


emersion:
Gentle ping

mholt:
Thanks; been heads-down in CertMagic code this week. Will get around to this!

mholt (approved):
LGTM. I just realized we don't need/use this in Caddy because we replace the entire cache on every config reload, rather than computing a delta and applying only the difference. But I guess as a library method this makes sense in the general case.

Thanks!

Edit: And no, storage cleanups typically happen in the background through `CleanStorage` :https://pkg.go.dev/github.com/caddyserver/certmagic#CleanStorage - it removes expired assets.

Issue #104, opened 2020-10-10
Custom HTTP Challenge Handler?


coolaj86:
I've got a special use case with a special problem (related to the go port of [telebit](https://telebit.cloud)).

I have a weird situation involving devices on private networks, tunneling, firewalls, and finicky DNS that the only solution that I can see is to **push HTTP challenge artifacts** to a trusted public-facing server and that presents them.

I've looked over the documentation and I can't seem to find a custom http challenge handler.

Does one exist? If not, would you be open to be adding one?

## Can't use DNS

Typically in this scenario, DNS challenges would be the correct choice, however, I have to support a DNS provider which has _wildly_ different propagation times between ns1, ns2, ns3, etc.

Even after pre-check validation completes, and after an additional custom wait of 45+ seconds, the Let's Encrypt Servers (which appear to be in globally diverse regions) can hit an NS server (also in globally diverse regions) to which the challenge has not yet propagated... and fail.

## Can't use ALPN

I can't use ALPN validation because the access to some of these devices is controlled by an IP/Port whitelist and the Let's Encrypt servers won't be on that whitelist.

> There isn’t and such a list won’t be made in the future too. Let’s Encrypt keeps the right to change the IP addresses used for authentication at will and won’t release lists of it for security reasons.

Ironically, port 80 is clear to use because the public-facing gateway doesn't forward any unencrypted traffic to the internal devices, so the gateway can capture all port 80 traffic indiscriminately.

mholt:
What do you mean by "custom http challenge handler?" Can you describe your situation in more detail? Right now my only understanding is that you have decided a solution but I don't even know what the problem is.

coolaj86:
I think you call it "Solver" and "Presenter" in your code. You've got one for DNS:

```go
import "github.com/libdns/cloudflare"

certmagic.DefaultACME.DNS01Solver = &certmagic.DNS01Solver{
	DNSProvider: cloudflare.Provider{
		APIToken: "topsecret",
	},
}
```

I want one for HTTP:

```go
import "github.com/xxx/http01"

certmagic.DefaultACME.HTTP01Solver = &certmagic.HTTP01Solver{
	HTTPProvider: http01.Provider{
		APIToken: "topsecret",
	},
}
```

mholt:
I see. Okay, will give it some thought!

coolaj86:
[...]

PR #108, opened 2020-11-03
allow custom HTTP-01 TLS-ALPN-01 challenge handlers

coolaj86:
Re: https://github.com/caddyserver/certmagic/issues/104

### DIFF
--- a/acmeclient.go
+++ b/acmeclient.go
@@ -144,7 +144,7 @@ func (am *ACMEManager) newACMEClient(ctx context.Context, useTestCA, interactive
 	// in situations where the default challenges would fail)
 	if am.DNS01Solver == nil {
 		// enable HTTP-01 challenge
-		if !am.DisableHTTPChallenge {
+		if !am.DisableHTTPChallenge && am.HTTP01Solver == nil {
 			useHTTPPort := HTTPChallengePort
 			if HTTPPort > 0 && HTTPPort != HTTPChallengePort {
 				useHTTPPort = HTTPPort
@@ -152,7 +152,7 @@ func (am *ACMEManager) newACMEClient(ctx context.Context, useTestCA, interactive
 			if am.AltHTTPPort > 0 {
 				useHTTPPort = am.AltHTTPPort
 			}
-			client.ChallengeSolvers[acme.ChallengeTypeHTTP01] = distributedSolver{
+			am.HTTP01Solver = distributedSolver{
 				acmeManager: am,
 				solver: &httpSolver{
 					acmeManager: am,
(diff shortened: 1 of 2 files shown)


francislavoie:
If HTTP-01 is customizable, shouldn't ALPN be as well?

coolaj86:
@francislavoie updated

mholt (approved):
I actually think this is pretty good! Thank you!

Sorry it took me a while. I worked through a backlog of about 110 notifications and now I'm down to 4. :partying_face:

mholt:
Oh... actually when you have a chance can you resolve the merge conflict? I would do it on the web interface but I've got another thing to attend to right now and I don't want to break things :sweat_smile:  

While you're at it I'm going to add one or two comments that are just nits.

acmemanager.go
```diff
@@ -73,6 +73,12 @@ type ACMEManager struct {
 	// from this package
 	DNS01Solver acmez.Solver
 
+	// The solver for the http-01 challenge
+	HTTP01Solver acmez.Solver
+
+	// The solver for the tls-alpn-01 challenge
+	TLSALPN01Solver acmez.Solver
```
mholt, line 80: Two things:

1. Can we move these above DNS01Solver? Sounds dumb, and it is dumb, but I usually see the order being: HTTP, TLSALPN, then DNS last; I just like to keep it consistent.

2. Can the godoc comments on these fields explain that these are _optional_ and that the default solvers work well for most people, including supporting distributed solving.

acmeclient.go
```diff
@@ -144,15 +144,15 @@ func (am *ACMEManager) newACMEClient(ctx context.Context, useTestCA, interactive
...
+		if !am.DisableHTTPChallenge && am.HTTP01Solver == nil {
 			useHTTPPort := HTTPChallengePort
 			if HTTPPort > 0 && HTTPPort != HTTPChallengePort {
 				useHTTPPort = HTTPPort
 			}
 			if am.AltHTTPPort > 0 {
 				useHTTPPort = am.AltHTTPPort
 			}
-			client.ChallengeSolvers[acme.ChallengeTypeHTTP01] = distributedSolver{
+			am.HTTP01Solver = distributedSolver{
```
mholt, line 155: On second thought; instead of setting the exported field to a non-nil value here, can we set a local variable instead? I just want to avoid drastic changes to the `am` struct, it might surprise users after a method call that suddenly this field is non-nil.

(Same below)

mholt:
@coolaj86 I think a better solution exists now, with `GetACMEChallenge()` - https://github.com/caddyserver/certmagic/commit/7271e2f615920738553778ee531d22196191210c -- so in your HTTP handler, for example, you can get the challenge information from CertMagic and solve it yourself, if you want to.

mholt:
Going to close this due to inactivity but also because I am pretty sure what wants to be done is possible now as per the above comment. Feel free to let me know if I'm missing something, we can continue discussion + reopen if needed!

PR #109, opened 2020-11-05
Implement multiple issuer support

mholt:
This change refactors Config.Issuer to be Config.Issuers, an array of
issuers. Each Issuer will be tried in turn until one succeeds. During
retries, each attempt will try each configured Issuer. When loading
certs from storage, CertMagic will look in each Issuer's storage
location for a qualifying asset. If multiple Issuers have one in storage
then the most-recently-issued cert will be selected.

This is a breaking change in that Config now accepts a slice of Issuers
rather than a single Issuer. The Revoker field is removed, as supporting
it is optional anyway. If the Issuer is also a Revoker, it can be used
implicitly to revoke certificates.

Also added a const for ZeroSSL's ACME endpoint.

Also, bug fixes for distributed solving of TLS-ALPN challenge:

The type assertion in handshake.go was problematic since there's no
guarantee that an ACME issuer would be a concrete ACMEManager type.

Refactored the code to accept IssuerKey values generally, rather than
specific ACMEManager values only.

This fixes solving tls-alpn challenges in distributed settings.

More cleanup can be done, another time.

### DIFF
--- a/account.go
+++ b/account.go
@@ -242,8 +242,12 @@ func (am *ACMEManager) askUserAgreement(agreementURL string) bool {
 	return answer == "y" || answer == "yes"
 }
 
+func storageKeyACMECAPrefix(issuerKey string) string {
+	return path.Join(prefixACME, StorageKeys.Safe(issuerKey))
+}
+
 func (am *ACMEManager) storageKeyCAPrefix(caURL string) string {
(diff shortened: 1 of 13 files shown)


mholt:
Going to merge this now, to get people to start using it and hopefully get any bug reports in sooner rather than later.

Issue #113, opened 2020-12-31
missing method AppendRecords using Readme.md example code


cameronelliott:
Update:
I have submitted the worlds smallest PR that fixes the Readme example to compile.
Go 1.15.6 (maybe Go used to compile the example code in older versions???)

I got this issue with both DuckDns and Cloudflare, trying to use the example from the Readme.
My apologies if this is a mistake on my end.

## What version of the package are you using?
github.com/caddyserver/certmagic v0.12.0
github.com/libdns/duckdns v0.0.0-20201229185916-cd405ff644b9
github.com/libdns/cloudflare v0.0.0-20200528144945-97886e7873b1
.../go/pkg/mod/github.com/libdns/libdns@v0.1.0/libdns.go

## What are you trying to do?
Use certmagic example from Certmagic readme:
```
import "github.com/libdns/cloudflare"

certmagic.DefaultACME.DNS01Solver = &certmagic.DNS01Solver{
	DNSProvider: cloudflare.Provider{
		APIToken: "topsecret",
	},
}
```

This code results in this error:
`cannot use (cloudflare.Provider literal) (value of type cloudflare.Provider) as certmagic.ACMEDNSProvider value in struct literal: missing method AppendRecords (AppendRecords has pointer receiver)compiler`

## What steps did you take?
Just tried to compile example from Readme.md
I tried 'go get' all the related packages to update them.
I also tried 'go get -u' to see if that helped.
I looked for an issue with AppendRecords at libdns/libdns.
I can see the diff between the signatures, but not sure how to fix the issue.

## What did you expect to happen, and what actually happened instead?
I expected the code to compile without errors.

## How do you think this should be fixed?
I am not clear what the solution is. 

## Bonus: What do you use CertMagic for, and do you find it useful?
I'm using Certmagic as part of an SFU for WebRTC. Certmagic certainly rocks!
[I also used it once to add SSL/TLS/HTTP to a [WAMP server](https://wamp-proto.org/)

mholt:
(Thanks for the PR.) Indeed, this is a mistake, it should be a pointer because it has a mutex.

PR #114, opened 2020-12-31
fix dns solver example code to compile okay

cameronelliott:
As reported in #113, the example code for implementing a DNS provider does not compile,
and in my opinion results in a cryptic message. (maybe everybody else figures it out? )

Nonetheless, I thought this fix would be a nice contribution to help others avoid having to try to figure out
that message.


mholt (approved):
This is correct, thank you!

PR #119, opened 2021-02-01
Added go mod init to the Readme

cryptopragmatic:


mholt:
Ah, I see why you were having a problem now.

Regarding the second commit (the one that passed CI tests and has no changes to the code), it's not really our job to teach how Go works, the user should already know how to use the `go` command. Doing so would get tedious to maintain and is out of scope for a library's readme, because pretty much all Go projects work the same way.

Please review the basic Go documentation:

- https://golang.org/doc/tutorial/getting-started
- https://golang.org/doc/tutorial/create-module
- https://golang.org/doc/code.html

All of these tutorials explain how to write Go code and will be of help to you.

As for the other two commits, they are broken, so I will be closing this PR. Thanks for understanding.

PR #125, opened 2021-03-26
Update cpuid

mohd-akram:


mholt:
IIRC we had to defer updating this because of a fatal crash on some ARM platforms. https://github.com/caddyserver/caddy/issues/3537

Do you know if that is fixed in the latest version? (Haven't had a chance to check.)

francislavoie:
ARM support was disabled/reverted in v1.3.1 then re-introduced with major changes in v2.0.0 https://github.com/klauspost/cpuid/releases

mholt (approved):
Cool, in that case, thanks for the patch! Apparently works on new M1 macs too.

Issue #121, opened 2021-03-15
Allow setting filestorage umask


SuperQ:
## What would you like to have changed?

The filestorage method currently hard codes the permissions to user-only access to the cert/key files generated. 

Allow passing permission, or at a minimum "allow group read" to the cert/key dirs/files.

## Why is this feature a useful, necessary, and/or important addition to this project?

This would make it possible for other services to use the TLS keys to be re-used.

francislavoie:
The file writes happen here: https://github.com/caddyserver/certmagic/blob/7550222c4a6a390cb0b91426d059f1ec5bda59b3/filestorage.go#L51

SuperQ:
@francislavoie Yup, that's how I found it was hard coded. :smile: 

The question is, should we allow directly changing the mask, or have an opinionated boolena option for "allow group read"?

mholt:
Changing that mask is a bit scary. I'd be more comfortable if something external did it, with intention.

francislavoie:
I think this might fall under "something an event system in Caddy could solve", i.e. on cert write, a event handler that triggers the file permission to be changed, if you need it to be... I guess.

francislavoie:
Alternatively, fork the `filestorage` module (maybe like `unsafe_filestorage`) with a wider default permission, or configurable permissions. :man_shrugging:

mholt:
Yeah, something like that. I just don't want people changing this without it being a part of another system that actually uses those permissions -- otherwise you get code or config that's copied and pasted that's insecure for no good reason. I think loosening restrictions needs to be done in the system that needs them loosened, rather than in the system that needs to be secure.

SuperQ:
@mholt Changing the umask to allow group access is a completely standard way to do things, and not scary at all.

SuperQ:
>  I think loosening restrictions needs to be done in the system that needs them loosened

The problem is, UNIX file mask permissions don't work this way. The file owner (Caddy/certmagic) needs to allow access. Without allowing things to change, there's no way for an external system to manage access.

I would be perfectly happy with hard-coding a change from `0600` to 0640`, so that the managing system that creates the caddy user/group can then manage what users are allowed to read from the caddy group. I'd also be happy with a bool that says "Allow group read". We don't need to get fancy. Just enough such that group read permissions are allowed.

SuperQ:
[...]

PR #128, opened 2021-04-20
Add option to allow group read in FileStorage

SuperQ:
Add a flag to FileStorage to allow group read access to the files
created. This allows for other systems to be granted group access to the
certs/keys created.

Fixes: https://github.com/caddyserver/certmagic/issues/121

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/filestorage.go
+++ b/filestorage.go
@@ -32,7 +32,8 @@ import (
 // directory. It is used to get file paths in a consistent,
 // cross-platform way or persisting ACME assets on the file system.
 type FileStorage struct {
-	Path string
+	Path           string
+	AllowGroupRead bool
 }
 
 // Exists returns true if key exists in fs.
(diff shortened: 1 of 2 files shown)


mholt:
Thanks, but as discussed in the issue, I don't think this is a good way to approach this; (also it does not compile).

SuperQ:
Without this, there is no way to use the keys that Caddy downloads with other software. Use UNIX groups is the standard way for this to happen. There is no option, other than having to disable Caddy's ACME support and fully going with an external tool.

mholt:
The user and group you run Caddy as is a system admin concern unfortunately, not a Caddy concern -- it should be done by other tooling.

SuperQ:
Yes, it absolutely is a system concern. That's fine. But the files Caddy creates need to either respect the umask the system presents, for example [in systemd, you can set the umask for a service](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#UMask=). Or the application needs to be told what umask to write files as.

Without one of those two options, I can't grant group permissions without race conditions to make this work.

Please reconsider your stance on this. This is a standard way for people to manage these files. It worked just fine on Caddy v1, where files respected the group permissions. We need a way to be able to control Caddy's file creations.

PR #131, opened 2021-05-28
Prioritize the configured resolvers

kmpm:
Make configured resolvers important and do not allow any fallback or other ones being used.
This solves some issues when for example using a split-head DNS and certmagic falls back to using /etc/resolve.conf.
It forces the use of only the configured resolvers and nothing else.

__ Caddyfile__
```
resolvers ns1.loopia.se !important
# or
resolvers ns1.loopia.se!important
```

__JSON__ 
```json
"resolvers": [
    "ns1.loopia.se",
    "!important"
  ]
```
alternatively
```json
"resolvers": [
    "ns1.loopia.se!important"
  ]
```

### DIFF
--- a/dnsutil.go
+++ b/dnsutil.go
@@ -312,12 +312,36 @@ func updateDomainWithCName(r *dns.Msg, fqdn string) string {
 	return fqdn
 }
 
+// isImportant is used to check the custom nameservers for
+// a suffix '!important'. If found, the suffix is removed from 
+// the records and the bool result is set to true
+func isImportant(custom []string) ([]string, bool) {
+	important := false
+	list := make([]string, len(custom))
+	for i, v := range custom {
+		if strings.HasSuffix(v, "!important") {
+			important = true
+			list[i] = strings.TrimSuffix(v, "!important")
+		} else {
+			list[i] = v
+		}
+	}
+	return list, important
+}
+
 // recursiveNameservers are used to pre-check DNS propagation. It
 // prepends user-configured nameservers (custom) to the defaults
 // obtained from resolv.conf and defaultNameservers and ensures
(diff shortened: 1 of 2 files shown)


dnsutil.go
```diff
@@ -313,12 +313,19 @@ func updateDomainWithCName(r *dns.Msg, fqdn string) string {
...
-	servers := append(custom, systemOrDefaultNameservers(defaultResolvConf, defaultNameservers)...)
+	var servers []string
+	if len(custom) == 0 {
+		servers = systemOrDefaultNameservers(defaultResolvConf, defaultNameservers)
+	} else {
+		servers = make([]string, len(custom))
+		copy(servers, custom)
+	}
 	populateNameserverPorts(servers)
+	fmt.Printf("populated servers, ignoring resolve.conf: %v", servers)
```
mholt, line 328: Remove this; proper way to emit this in production would be a log, but we'd have to refactor for that. (not necessary for this PR IMO).

```suggestion
```

mholt (changes requested):
I think this is much better, thank you! Just one stray line to remove.

dnsutil.go
kmpm: fixed, sorry

mholt (approved):
Awesome, looks perfect now. Let's give this a shot. Thanks for the contribution!

mholt (approved):
Ah, thanks for the quick fix! Tests are passing now. :+1:

mholt:
Really appreciate your contribution. Feel free to participate again sometime!

kmpm:
It's been a VERY pleasant experience contributing to certmagic, caddy and caddy-dns. So you bet I'll be back.

PR #134, opened 2021-06-11
Refactor ObtainCert and RenewCert to allow force renews

mholt:
See https://github.com/caddyserver/caddy/issues/4191

### DIFF
--- a/config.go
+++ b/config.go
@@ -412,40 +412,31 @@ func (cfg *Config) Unmanage(domainNames []string) {
 	cfg.certCache.mu.Unlock()
 }
 
-// // ObtainCert obtains a certificate for name using cfg, as long
-// // as a certificate does not already exist in storage for that
-// // name. The name must qualify and cfg must be flagged as Managed.
-// // This function is a no-op if storage already has a certificate
-// // for name.
-// //
-// // It only obtains and stores certificates (and their keys),
-// // it does not load them into memory. If interactive is true,
-// // the user may be shown a prompt.
-// // TODO: consider moving interactive param into the Config struct,
-// // and maybe retry settings into the Config struct as well? (same for RenewCert)
-// func (cfg *Config) ObtainCert(ctx context.Context, name string, interactive bool) error {
-// 	if len(cfg.Issuers) == 0 {
-// 		return fmt.Errorf("no issuers configured; impossible to obtain or check for existing certificate in storage")
-// 	}
-// 	if cfg.storageHasCertResourcesAnyIssuer(name) {
-// 		return nil
-// 	}
-// 	// ensure storage is writeable and readable
-// 	// TODO: this is not necessary every time; should only perform check once every so often for each storage, which may require some global state...
-// 	err := cfg.checkStorage()
-// 	if err != nil {
-// 		return fmt.Errorf("failed storage check: %v - storage is probably misconfigured", err)
-// 	}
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -894,6 +938,29 @@ func (cfg *Config) storageHasCertResources(issuer Issuer, domain string) bool {
 		cfg.Storage.Exists(metaKey)
 }
 
+// deleteSiteAssets deletes the folder in storage containing the
+// certificate, private key, and metadata file for domain from the
+// issuer with the given issuer key.
+func (cfg *Config) deleteSiteAssets(issuerKey, domain string) error {
```
francislavoie, line 944: FWIW, I think this would be useful to expose, could allow for implementing a CLI command to wipe out assets for a site from CLI in Caddy (or via the admin API maybe).

mholt: I like that idea. Will need to think about a good way to do this.

francislavoie: Maybe 3rd param `fullWipe bool` or something?

maintain.go
```diff
@@ -368,18 +371,28 @@ func (certCache *Cache) updateOCSPStaples(ctx context.Context) {
...
+		// If a managed certificate was revoked, we should attempt to replace it with a new one.
 		if cert.managed && ocspResp.Status == ocsp.Revoked && len(cert.Names) > 0 {
 			// if revoked for key compromise, we can't be sure whether the storage of
 			// the key is still safe; however, we KNOW the old key is not safe, and we
 			// can only hope by the time of revocation that storage has been secured;
 			// key management is not something we want to get into, but in this case
 			// it seems prudent to replace the key
+			var obtainInsteadOfRenew bool
 			if ocspResp.RevocationReason == acme.ReasonKeyCompromise {
-				// TODO: delete site assets, then call Obtain instead
```
francislavoie: Removing TODOs

image

config.go
mholt: This'll have to come in another feature/PR.

maintain.go
```diff
@@ -66,6 +67,11 @@ func (certCache *Cache) maintainAssets(panicCount int) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// since checking certs for renewal is done when they're loaded, we don't need
+	// to do that until the first tick; but OCSP staples aren't necessarily refreshed
+	// at load-time, so we do this real quick before the first tick (see #4191)
+	certCache.updateOCSPStaples(ctx)
```
francislavoie, line 73: How will this behave with the recursion with panics? Will it make this happen too often in a tight loop if something goes wrong?

mholt: Good question.

I guess; but I've yet to see that happen more than once ever, and it's a pretty lightweight function (I am not even sure there's any I/O in the nominal case). The panic recursion only happens at most 10 times.

francislavoie (approved):
I don't have the strongest grasp on this stuff but it looks pretty good to me 👍

mholt:
Good thing I tested this to some extent. :stuck_out_tongue:  I had broken the regular `Obtain()` flow, but now I feel pretty confident, as I went through Obtain, Renew flows without problems, and simulated OCSP revocation and it seemed that certificate replacement worked well, in cases of key compromise (where a new key is generated) or otherwise (reusing the same key) -- a new feature btw.

PR #139, opened 2021-06-25
Fix TLS-ALPN-01 challenge for IP Identifiers

hslatman:


mholt:
Thanks, I'll take a look at this soon!

solvers.go
```diff
@@ -648,6 +651,18 @@ type Challenge struct {
 	data interface{}
 }
 
+// challengeKey determines the key for a challenge Identifier
```
mholt, line 654:
```suggestion
// challengeKey returns the map key for a given challenge; it is the identifier
// unless it is an IP address using the TLS-ALPN challenge.
```

solvers.go
```diff
@@ -648,6 +651,18 @@ type Challenge struct {
...
+func challengeKey(chal acme.Challenge) string {
+	key := chal.Identifier.Value
+	if chal.Type == acme.ChallengeTypeTLSALPN01 && chal.Identifier.Type == "ip" {
+		k, err := dns.ReverseAddr(chal.Identifier.Value)
+		if err == nil {
+			key = k[:len(k)-1] // strip off .
+		}
+	}
+	return key
```
mholt, line 663: This might just be me nit-picking, but I feel like this flows a little better. :smile: 

```suggestion
	if chal.Type == acme.ChallengeTypeTLSALPN01 && chal.Identifier.Type == "ip" {
		reversed, err := dns.ReverseAddr(chal.Identifier.Value)
		if err == nil {
			return reversed[:len(reversed)-1] // strip off '.'
		}
	}
	return chal.Identifier.Value
```

mholt (changes requested):
Sorry this took a while. Looks good, pretty straightforward change.

Just to confirm, is `chal.Type == acme.ChallengeTypeTLSALPN01 && chal.Identifier.Type == "ip"` correct? We only reverse the IP for the TLS-ALPN challenge, but not for the HTTP challenge?

hslatman:
> Just to confirm, is `chal.Type == acme.ChallengeTypeTLSALPN01 && chal.Identifier.Type == "ip"` correct? We only reverse the IP for the TLS-ALPN challenge, but not for the HTTP challenge?

Yes, if I understand the [RFC](https://datatracker.ietf.org/doc/html/rfc8738) correctly:

> As [RFC6066] does not permit IP addresses to be used in the Server Name Indication (SNI) extension HostName field, the server MUST instead use the IN-ADDR.ARPA [RFC1034] or IP6.ARPA [RFC3596] reverse mapping of the IP address as the HostName field value instead of the IP address string representation itself.

For the HTTP challenge it mentions the following:

> For the "http-01" challenge, the Host header field MUST be set to the IP address being used for validation per [RFC7230].

Now that I had another look at this I need to find out how caching of the challenge certificate works. Actually, deletion of the cached challenge certificate here: https://github.com/caddyserver/certmagic/blob/647f27cf265e6f72b6f043ac52ba34f01bb5da56/solvers.go#L219

That's the only case that still uses the IP instead of the reversed one and I'm suspicious that this cert may not be properly removed after the challenge is solved if a different key is used there. I may be missing something, but the challenge cert never seems to be inserted into the certificate cache and to only exist in the active challenges map? If that's the case, the cleanup doesn't seem necessary, but I may be on a wrong path here.

hslatman:
I did find a reference to RFC6066 here:

https://github.com/caddyserver/certmagic/blob/647f27cf265e6f72b6f043ac52ba34f01bb5da56/handshake.go#L129-L144

That part only retrieves the certificate though, so it seems a coincidence.

solvers.go
hslatman: 👍

mholt:
@hslatman Ah, great find. Indeed, legacy versions of CertMagic used to store the challenge certificate directly into the certificate cache. Now, we only store the challenge info in the configured storage and generate the certificate on-the-fly during the handshake that needs it. This allows us to cleanly distribute the challenge and makes solving of the challenge agnostic of the challenge type used.

So that `delete()` call you highlighted can actually be removed. And I could _probably_ remove the code in handshake.go that checks for an exact match of the cert hash from the SNI value; AFAIK we haven't used that functionality since the TLS-SNI days.

hslatman:
I've removed the delete from Cleanup. The other part feels more appropriate for a separate issue, indeed.

mholt (approved):
Excellent, thanks so much for this improvement. Really appreciate it!

PR #142, opened 2021-08-16
Load certs from storage if cache is full

mholt:
Should fix "no certificate available" errors.

https://caddy.community/t/error-tls-alert-internal-error-592-again/13272/9?u=matt

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -214,9 +214,15 @@ func (cfg *Config) getCertDuringHandshake(hello *tls.ClientHelloInfo, loadIfNece
 
 	name := cfg.getNameFromClientHello(hello)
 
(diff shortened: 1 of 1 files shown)


francislavoie (approved):
As an aside, I think we should also look into making the cache capacity configuration via Caddyfile (it is JSON configurable currently, see `cacheOpts.Capacity` in Caddy)

mholt:
There was an edge case where this patch wasn't quite sufficient... a replacement cert might itself be evicted from the cache, leaving the cache less-than-100%-full. We now allow for the cache to be only 90% full or higher: 88b8609b4d8fb8c7b1865322602606b85240c80d

PR #150, opened 2021-10-31
Provide more detailed information in certificate events

francislavoie:
The `cert_*` events were only providing the domain. This was pretty limited, because not much can be done with it without also knowing the issuer, since that affects where the cert is stored.

Providing both the `issuerKey` and `certKey` makes it possible for the event listener to call `StorageKeys.SiteCert(issuerKey, certKey)` to get the actual location of the cert (and similarly the private key). Particularly useful in situations where the user wants to copy the cert/key to some location for another app to use the cert that is being managed.

A similar change could probably also be done for `cached_managed_cert` and `cached_unmanaged_cert` events to provide more detail, because `cert.Names` string slice doesn't seem that useful, can't do much with that alone. But I'm not sure I understand the usefulness of those events, so meh. Caching seems like pretty much an internal implementation detail of certmagic.

Note: This is a breaking change for any users of certmagic that use `OnEvent`, but only because we've wrapped the existing information with a new type, to provide _more_ information. Shouldn't harm anyone.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1107,6 +1107,7 @@ type OCSPConfig struct {
 
 // CertificateEventData contains contextual information for
 // an obtained, renewed or revoked certificate.
+// EXPERIMENTAL: subject to change.
 type CertificateEventData struct {
 	// Domain or subject name of the certificate.
 	Name string
--- a/config.go
+++ b/config.go
@@ -573,10 +573,10 @@ func (cfg *Config) obtainCert(ctx context.Context, name string, interactive bool
 			return fmt.Errorf("[%s] Obtain: saving assets: %v", name, err)
 		}
 
-		cfg.emit("cert_obtained", CertEventData{
-			Name:      name,
-			IssuerKey: issuerUsed.IssuerKey(),
-			CertKey:   certRes.NamesKey(),
+		cfg.emit("cert_obtained", CertificateEventData{
+			Name:       name,
+			IssuerKey:  issuerUsed.IssuerKey(),
+			StorageKey: certRes.NamesKey(),
 		})
 
 		if log != nil {
@@ -793,10 +793,10 @@ func (cfg *Config) renewCert(ctx context.Context, name string, force, interactiv
 			return fmt.Errorf("[%s] Renew: saving assets: %v", name, err)
 		}
 
-		cfg.emit("cert_renewed", CertEventData{
-			Name:      name,
-			IssuerKey: issuerUsed.IssuerKey(),
-			CertKey:   certRes.NamesKey(),
+		cfg.emit("cert_renewed", CertificateEventData{
+			Name:       name,
+			IssuerKey:  issuerUsed.IssuerKey(),
+			StorageKey: certRes.NamesKey(),
 		})
 
 		if log != nil {
@@ -876,10 +876,10 @@ func (cfg *Config) RevokeCert(ctx context.Context, domain string, reason int, in
 			return fmt.Errorf("issuer %d (%s): %v", i, issuerKey, err)
 		}
 
-		cfg.emit("cert_revoked", CertEventData{
-			Name:      domain,
-			IssuerKey: issuerKey,
-			CertKey:   certRes.NamesKey(),
+		cfg.emit("cert_revoked", CertificateEventData{
+			Name:       domain,
+			IssuerKey:  issuerKey,
+			StorageKey: certRes.NamesKey(),
 		})
 
 		err = cfg.deleteSiteAssets(ctx, issuerKey, domain)
@@ -1105,17 +1105,17 @@ type OCSPConfig struct {
 	ResponderOverrides map[string]string
 }
 
-// CertEventData contains contextual information for
+// CertificateEventData contains contextual information for
 // an obtained, renewed or revoked certificate.
-type CertEventData struct {
+type CertificateEventData struct {
 	// Domain or subject name of the certificate.
 	Name string
 
 	// Storage key for the issuer used for this certificate.
 	IssuerKey string
 
-	// Storage key for the certificate by its name.
-	CertKey string
+	// Location in storage at which the certificate could be found.
+	StorageKey string
 }
 
 // certIssueLockOp is the name of the operation used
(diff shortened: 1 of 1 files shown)


mholt:
I also don't know of any integrations with events yet. I agree it could be more helpful though. Maybe instead of passing the names of things, we could pass the actual values (the actual Issuer and CertificateResource, instead of just their serializations).

francislavoie:
Well, `Issuer` is just an interface, and the only thing useful on it is `IssuerKey()` (because no consumer of the event should ever call `Issue()`, doesn't make sense), and the underlying impl `ACMEManager` has quite a lot of not-so-useful things (basically config, which is redundant to pass in the event data).

`CertificateResource` is more useful though, because it comes with the actual PEM of the cert/key pair.

Also, the event listener might actually want to handle the event in a goroutine. Should we make a copy of `CertificateResource` first? I'm still not super sure how we need to handle the data in case of concurrency.

mholt:
Hmm, copies are probably a good idea for something like this even without concurrency, but I don't love the performance implications.

What if we just say "Event handlers are synchronous - if you use goroutines you must copy the data first" in the docs...

francislavoie:
I guess, that'd mean we'd still have to always do the copy in Caddy before dispatching it (see my WIP in https://github.com/francislavoie/caddy/commit/2110fd84e0c9c4137d99709456a806a0ac4b9ad8)

mholt:
> I guess, that'd mean we'd still have to always do the copy in Caddy before dispatching it

That's what I want to avoid. If the handler isn't a quick, read-only thing, then I think it's fair to ask the consumer to copy the data first. So only if necessary.

francislavoie:
What I mean is we don't have that choice, we don't know what listeners may be registered. There's no way to _not_ do the copy in Caddy before dispatching it, I don't think.

mholt:
It doesn't matter though. We just document to consumers of the event system that they need to copy the data if they're going to use concurrency.

config.go
```diff
@@ -1077,6 +1089,19 @@ type OCSPConfig struct {
 	ResponderOverrides map[string]string
 }
 
+// CertificateEventData contains contextual information for
+// an obtained, renewed or revoked certificate.
+type CertificateEventData struct {
```
mholt, line 1094:
```suggestion
type CertificateEvent struct {
```

config.go
```diff
@@ -1077,6 +1089,19 @@ type OCSPConfig struct {
...
+	// Domain or subject name of the certificate.
+	Name string
```
mholt, line 1096:
```suggestion
	Names []string
```

It's true there is only 1 name on managed certificates, but certificates in general can have multiple subjects.

config.go
```diff
@@ -1077,6 +1089,19 @@ type OCSPConfig struct {
...
+
+	// Storage key for the issuer used for this certificate.
+	IssuerKey string
```
mholt, line 1099:
```suggestion
	// The issuer used to manage this certificate (if applicable).
	Issuer Issuer
```

Might as well include the actual value, then the user can get the key as well as do type assertions, etc; more useful.

config.go
```diff
@@ -1077,6 +1089,19 @@ type OCSPConfig struct {
...
+
+	// Storage key for the certificate by its name.
+	CertificateKey string
```
mholt, line 1102:
```suggestion
	// Location in storage at which the certificate could be found.
	StorageKey string
```

mholt (changes requested):
Thanks for enhancing this Francis. What if we also add the Config to the `CertificateEvent` struct?

And actually, we might even be able to replace the Name and Issuer fields with a single CertificateResource field. What do you think?

config.go
francislavoie: As I said earlier, this would require a copy, because this will likely get fired via a goroutine. It's not safe to pass the actual issuer. See https://github.com/caddyserver/certmagic/pull/150#issuecomment-956557862

config.go
francislavoie: This isn't the actual event, the event is `cert_obtained` etc. This is the data attached to the event.

config.go
francislavoie: Hmm, I guess, but all the emissions currently are done in the context of a single name.

config.go
mholt: I think it's safe if it's read-only, yeah? And while it's true we don't think people should be calling Issue(), it's possible that type-assertions could be useful to callers.

In any case, I don't care too strongly on this. Leaving it as a string is fine if that's best.

config.go
mholt: Oh fine. What about `EventData` and then we can add fields that might just be empty depending on the kind of event?

config.go
mholt: Managed certificates aren't multi-SAN, but any manually-provided certificate can have multiple SANs. We don't want to limit events to only managed certificates (for example, think events for OCSP staples, etc).

config.go
mholt: True. I wonder if we should change the `emit()` signature to simply take an `Event` struct where the `Name string` is a required field, and the `Data` can be type-asserted... that way the name and data stay together.

Oh and a timestamp might not be a bad idea.

config.go
mholt: I don't know that we can guarantee we'd never have certificate-oriented events for unmanaged certificates though. I do feel like []string is crucial here.

francislavoie: What if we called it `ManagedCertificateEventData` or something?

config.go
mholt: Ah right. And it's true, maybe Timestamp isn't super useful. Just seems like a prudent thing to have to document when an event took place in case they want to store or process it or something. Or if event handling is blocked/lagging behind.

mholt:
I think I'm good with merging this, but can we mark it as experimental (at least the `CertificateEventData` struct)? Just so we can more easily change it later if needed. It'll be hard to make further improvements to this change without real-world experience in production and feedback from users.

francislavoie:
Yep 👍

mholt (approved):
Thanks, let's give this a try and iterate on it with time and feedback!

(GitHub Actions is down _again_ so I will just merge.)

PR #153, opened 2021-11-23
duplicate OnDemand default value assign check

rjbasitali:
The `newWithCache` func inside `config.go` has duplicate check and assign statement for `cfg.OnDemand`.


mholt (approved):
Ah, thanks! Never noticed this. :)

PR #155, opened 2021-12-14
Propagate context in the Storage interface methods

### LINKED ISSUE #154: Add context propagation to the Storage interface

The `Storage` interfaces do not support context propagation. Only the `Lock` method does.  This would require changing the method signature for the `Storage` interface: 
```go
// Implementations of Storage must be safe for concurrent use.
type Storage interface {
	// Locker provides atomic synchronization
	// operations, making Storage safe to share.
	Locker

	// Store puts value at key.
	Store(ctx context.Context, key string, value []byte) error

	// Load retrieves the value at key.
	Load(ctx context.Context, key string) ([]byte, error)

	// Delete deletes key. An error should be
	// returned only if the key still exists
	// when the method returns.
	Delete(ctx context.Context, key string) error

	// Exists returns true if the key exists
	// and there was no error checking.
	Exists(ctx context.Context, key string) bool

	// List returns all keys that match prefix.
	// If recursive is true, non-terminal keys
	// will be enumerated (i.e. "directories"
	// should be walked); otherwise, only keys
	// prefixed exactly by prefix will be listed.
	List(ctx context.Context, prefix string, recursive bool) ([]string, error)

	// Stat returns information about key.
	Stat(ctx context.Context, key string) (KeyInfo, error)
}
```

Propagating the context would be useful when using a remote storage (mainly for context cancellation). For example with the s3 storage:

https://github.com/ss098/certmagic-s3/blob/efad3a48d4edfbe2fee64e9c4d04195603822b07/s3.go#L144

hairyhenderson:
Fixes #154

This adds a `context.Context` as the first argument to each method in the `Storage` and `Locker` interfaces, and propagates existing contexts as much as possible. All uses of `context.TODO()` are replaced.

Instead of changing the `HTTPS` function's signature, I've added an `HTTPSWithContext` as it's a breaking change and that's likely the most-used function. I'm not sure if I should add a `TLSWithContext` and `ListenWithContext` too, or change those, or leave them as-is... 🤔

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


hairyhenderson:
🤦‍♂️ I didn't catch that we needed to support Go 1.14, and I also didn't realize that `ClientHelloInfo.Context` is new [as of Go 1.17](https://go.dev/doc/go1.17#crypto/tls) - so I probably can't use that. I'll circle back and fix that up.

francislavoie:
Caddy is 1.16+ so we could bump to that maybe, but yeah, not 1.17 minimum yet.

mholt:
Thanks for this, Dave.

To be honest, I'm fine with breaking changes in CertMagic still as we're pre-1.0, and I'd rather do it now rather than have messy "WithContext" functions later. So I'd just go ahead and add the context where it is needed, including those top-level functions.

Will start reviewing this shortly!

handshake.go
```diff
@@ -614,16 +616,18 @@ func (cfg *Config) renewDynamicCertificate(hello *tls.ClientHelloInfo, currentCe
...
 	}
 
+	ctx := hello.Context()
+
 	// if the certificate hasn't expired, we can serve what we have and renew in the background
 	if timeLeft > 0 {
-		// TODO: get a proper context; we use one with timeout because retries are enabled because interactive is false
-		ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Minute)
+		// we use this timeout because retries are enabled because interactive is false
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
chaudyg, line 624: 🤔 I am wondering if we should keep this one set to `ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)`? `go renewAndReload(ctx, cancel)` is a background task. Do we want to cancel the task if the request gets canceled?

hairyhenderson: This is a good question. My understanding is that this is all done within the context of the `*tls.ClientHelloInfo` - _currently_ this uses the actual `context.Context` on that struct, but that's a Go 1.17+ thing, so I'll be reworking this. 

Either way I think it still makes sense to propagate _a_ context here. It'll end up being fairly high-level/long-lived context, and we definitely want background tasks spawned here to be cancelled by it.

francislavoie: I think this context is to make On-Demand TLS work (i.e. fetch a cert during the TLS handshake if the domain is eligible). Something I was mentioning to Matt recently is that it would be ideal to transfer this process to the background instead on timeout in case the issuance is slow (shouldn't be slow but it sometimes is if the ACME issuer is having problems).

I'm not saying that something different needs to be done here, but it's something to consider.

mholt: Yeah, so this context is tricky because it is a background task even though it is initiated by a TLS handshake. It's unclear to me what the proper context should be. The duration of the handshake doesn't make sense because it's likely that the goroutine would be cancelled too early.

This one might need to stay `context.TODO()` for now until we can find the right value here.

@francislavoie 

> I think this context is to make On-Demand TLS work (i.e. fetch a cert during the TLS handshake if the domain is eligible). Something I was mentioning to Matt recently is that it would be ideal to transfer this process to the background instead on timeout in case the issuance is slow (shouldn't be slow but it sometimes is if the ACME issuer is having problems).

Not quite; this is on-demand TLS, but specifically the case when the certificate we're serving is expiring soon. We already have the cert, but it needs to be renewed (and has not yet expired). Thus we can serve it immediately and renew it in the background for performance reasons (no need to block the handshake). This already is in the background so it's OK if issuance is slow -- but that's also why we can't tether it to the current handshake's context.

mholt:
@hairyhenderson Any interest in finishing up this PR? We'll want to coordinate submitting PRs to all the known storage implementations as a courtesy.

mholt (approved):
Ok, I went ahead and cleaned up the merge conflicts myself. It was a doozy, so I might have made some mistakes. This is obviously a breaking change, so next is to submit PRs to known storage implementations to help them transition. (This commit will be tagged for v0.16.0.)

Hope this works :crossed_fingers:

PR #156, opened 2021-12-15
Fix cacheAlmostFull calculation

skirsten:
This fixes a small bug that happened when the `cfg.certCache.options.Capacity` is 0 (unlimited).

I noticed that certificates I did not `Manage` or manually `Unmanage`'d got loaded from storage. This was unexpected and I traced it down to this bug that causes `cacheAlmostFull` to always be true, if the capacity is set to 0.

In the meantime, this can be mitigated by setting the capacity to `math.MaxInt32`.

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -265,7 +265,8 @@ func (cfg *Config) getCertDuringHandshake(hello *tls.ClientHelloInfo, loadIfNece
 	// perfectly full while still being able to load needed certs from storage.
 	// See https://caddy.community/t/error-tls-alert-internal-error-592-again/13272
 	// and caddyserver/caddy#4320.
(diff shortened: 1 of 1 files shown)


mholt:
Ohh, that's subtle. Thanks for the patch. I wonder if the same result could be achieved without the math.Inf approach (which feels a little hacky/unfortunate to me) -- what about something like this instead:

```
cacheAlmostFull := cacheCapacity > 0 && float64(cacheSize) >= (cacheCapacity * .9)
```

skirsten:
:+1: I updated the commit

mholt (approved):
Thank you! Looks great now.

PR #159, opened 2021-12-25
Remove useless active DNS challenges lock

skirsten:
Why remove this code?

- The lock is instance-local, so this will not work in a cluster setup
- > [You can have multiple TXT records in place for the same name. For instance, this might happen if you are validating a challenge for a wildcard and a non-wildcard certificate at the same time. However, you should make sure to clean up old TXT records, because if the response size gets too big Let’s Encrypt will start rejecting it.](https://letsencrypt.org/docs/challenge-types/#:~:text=You%20can%20have,start%20rejecting%20it.)
- The DNS code uses `libdns.RecordAppender` and not `libdns.RecordSetter` so the DNS interface is expected to support multiple records
- This unnecessarily hinders challenge solving performance

### DIFF
--- a/solvers.go
+++ b/solvers.go
@@ -261,13 +261,6 @@ func (s *DNS01Solver) Present(ctx context.Context, challenge acme.Challenge) err
 	dnsName := challenge.DNS01TXTRecordName()
 	keyAuth := challenge.DNS01KeyAuthorization()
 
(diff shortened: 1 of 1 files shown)


skirsten:
Just noticed that this was protecting against collisions in `s.txtRecords` in solvers.go. So this code makes more sense now.

PR #160, opened 2021-12-28
Support OverrideDomain is DNS01Solver

crccw:
CNAME can be used to delegate answering the chanllenge to another DNS
zone, in order to reduce the exposure of the DNS credential [1]. The
solver already follows CNAME when checking propagation against the
authoritative source, it should follow CNAME when setting the TXT record
as well.

[1] https://letsencrypt.org/docs/challenge-types/#dns-01-challenge

### DIFF
--- a/dnsutil.go
+++ b/dnsutil.go
@@ -214,10 +214,20 @@ func checkDNSPropagation(fqdn, value string, resolvers []string) (bool, error) {
 		fqdn += "."
 	}
 
+	// Initial attempt to resolve at the recursive NS
+	r, err := dnsQuery(fqdn, dns.TypeTXT, resolvers, true)
+	if err != nil {
+		return false, err
+	}
+
 	// TODO: make this configurable, maybe
 	// if !p.requireCompletePropagation {
 	// 	return true, nil
 	// }
+
+	if r.Rcode == dns.RcodeSuccess {
+		fqdn = updateDomainWithCName(r, fqdn)
+	}
 
 	authoritativeNss, err := lookupNameservers(fqdn, resolvers)
 	if err != nil {
--- a/dnsutil.go
+++ b/dnsutil.go
@@ -214,20 +214,10 @@ func checkDNSPropagation(fqdn, value string, resolvers []string) (bool, error) {
 		fqdn += "."
 	}
 
-	// Initial attempt to resolve at the recursive NS
-	r, err := dnsQuery(fqdn, dns.TypeTXT, resolvers, true)
-	if err != nil {
-		return false, err
-	}
-
 	// TODO: make this configurable, maybe
 	// if !p.requireCompletePropagation {
 	// 	return true, nil
 	// }
-
-	if r.Rcode == dns.RcodeSuccess {
-		fqdn = updateDomainWithCName(r, fqdn)
-	}
 
 	authoritativeNss, err := lookupNameservers(fqdn, resolvers)
 	if err != nil {
--- a/solvers.go
+++ b/solvers.go
@@ -252,6 +252,11 @@ type DNS01Solver struct {
 	// Preferred DNS resolver(s) to use when doing DNS lookups.
 	Resolvers []string
 
+	// Override the domain to set the TXT record on. This is
+	// to delegate the chanllenge to a different domain. Note
+	// that the solver doesn't follow CNAME/NS record.
+	OverrideDomain string
+
 	txtRecords   map[string]dnsPresentMemory // keyed by domain name
 	txtRecordsMu sync.Mutex
 }
@@ -259,6 +264,9 @@ type DNS01Solver struct {
 // Present creates the DNS TXT record for the given ACME challenge.
 func (s *DNS01Solver) Present(ctx context.Context, challenge acme.Challenge) error {
 	dnsName := challenge.DNS01TXTRecordName()
+	if s.OverrideDomain != "" {
+		dnsName = s.OverrideDomain
+	}
 	keyAuth := challenge.DNS01KeyAuthorization()
 
 	// multiple identifiers can have the same ACME challenge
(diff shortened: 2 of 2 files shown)


francislavoie:
This is related to https://github.com/caddyserver/caddy/issues/4071

So from my reading of the code, this would directly allow DNS records to directly affect where certmagic would attempt to mutate the TXT record.

I'm not sure that's proper. It is kinda cool to make it work dynamically like that, but it also kinda scares me, that it might have some unintended vector for abuse. But maybe I'm being paranoid. I just think we need to be careful.

I'd be more comfortable if there was an explicit domain override configurable instead of following the CNAME record. This is how it's set up in the DuckDNS plugin, so that is explicit at the config level. That config could be in certmagic itself instead, one layer higher (and Caddy would add an option for it).

Maybe if you could enumerate the ways this could be used, to prove it would be infeasible to abuse, it would feel more comfortable.

crccw:
Yes it would allow DNS record to affect where certmagic would attempt to mutate the TXT record.

It certainly makes sense to be paranoid when it comes to security. Though, I'm possibly being naive here, I don't feel it increases the exposure. In the end ACME checks the TXT record, which is independent from how the record is set. If a malicious party is able to put the TXT record there with this commit, they should be able to put it there without this commit too.

Being said, I like the idea of having a domain override configurable. It's simpler than following CNAME, and is more test friendly. If we do this, should we also stop following CNAME when checking propagation for consistency?

mholt:
Yeah, I think if it's specified in the config directly it should not be overrideable by any DNS records.

crccw:
Yes! Sorry for the delay, I'll update the PR in a day or two.

crccw:
Updated. I've tested the change together with caddyserver/caddy#4596. Let me know if there's a way to write a test against. I guess we need to mock a DNS server?

solvers.go
```diff
@@ -252,13 +252,21 @@ type DNS01Solver struct {
 	// Preferred DNS resolver(s) to use when doing DNS lookups.
 	Resolvers []string
 
+	// Override the domain to set the TXT record on. This is
+	// to delegate the chanllenge to a different domain. Note
```
mholt, line 256:
```suggestion
	// to delegate the challenge to a different domain. Note
```

dnsutil.go
```diff
@@ -214,21 +214,11 @@ func checkDNSPropagation(fqdn, value string, resolvers []string) (bool, error) {
...
-
 	// TODO: make this configurable, maybe
 	// if !p.requireCompletePropagation {
 	// 	return true, nil
 	// }
 
-	if r.Rcode == dns.RcodeSuccess {
-		fqdn = updateDomainWithCName(r, fqdn)
-	}
-
```
mholt, line 231: Are these removals necessary for getting the OverrideDomain feature to work? And does it still make sense for the commented block to remain if the surrounding code is now gone?

If these lines aren't required in general, I guess it's OK to remove them, but ideally I'd rather make them opt-in I think (configurable).

mholt (approved):
Ah, thanks for the update. LGTM now, let's give it a try. Really appreciate your contribution here! I get antsy trying to mess with this kind of code myself.

If there are new issues reported related to this, would you be able to help maintain this change?

crccw:
Yeah I'm happy to help.

IndeedNotJames:
Hi @crccw :)
What was the motivation behind https://github.com/caddyserver/certmagic/pull/160/commits/4be5af51f6ace5aeb48d33af9200c10569208be4#diff-0d423173eaa496408ee9e7d82ae770feaa3b860783ca6a155f6d7551cb5d4d14R337-R341

`checkDNSPropagation` and `checkAuthoritativeNss` don't seem to be actually interchangeable in that case here:
- `dnsName` has to be a fqdn in `checkAuthoritativeNss` but not in `checkDNSPropagation` https://github.com/caddyserver/certmagic/blob/9a56fcd4f9cc49b8068eab8674a9bd4468068ca3/dnsutil.go#L212-L215
- `resolvers` don't share a common format: `checkAuthoritativeNss` takes resolvers without port to add `:53` by itself, while `checkDNSPropagation` assumes the port already set.
- https://github.com/caddyserver/certmagic/blob/9a56fcd4f9cc49b8068eab8674a9bd4468068ca3/dnsutil.go#L243 needs to be `recursive=true` in my testings. (Though I am not sure what exactly that changes, but wouldn't resolve with `false` for me)

Using `checkDNSPropagation` in both cases (so dropping the above if/else again) seems to work just fine on the other hand.

mholt:
@crccw Was it intentional that the override domain always has to be prefixed with `_acme_challenge.`? Since that's part of the DNS-01 challenge spec, I am not sure why we would always want to make the user include that in the override domain...

francislavoie:
IMO we could adjust it to add that prefix to the configured override if it doesn't have the prefix already.

mholt:
Yeah, as long as there's not a legitimate reason to _not_ have that subdomain, I'm good with that. I can make that patch shortly.

crccw:
@mholt My memory has faded a little, but IIRC OverrideDomain can be anything. The user needs to create a CNAME record on the _acme_challenge.original.domain that points to the OverrideDomain. My understanding is that the CNAME should be, in some means, transparent to the ACME server; they just follow the CNAME chains when checking the TXT record on the _acme_challenge.original.domain.

Have you run into any issues if the override domain doesn't start with _acme_challenge?

mholt:
@crccw Thanks for replying! Yeah, we're trying to figure out, from this topic in the forum here: https://caddy.community/t/global-dns-challenge-and-dns-challenge-override-domain/16773

Is our patch (prepend `_acme-challenge.` if it doesn't exist) correct, do you think?

crccw:
I don't think so. In the post the author has a CNAME record from _acme-challenge.manage.clientsite.com > _acme-challenge.myexamplesite.com, therefore they need to set override_domain to the CNAME target _acme-challenge.myexamplesite.com. The CNAME target can be anything, e.g. _some-random-stuff.myexamplesite.com, in which case override_domain should be _some-random-stuff.myexamplesite.com accordingly.

mholt:
Oh I see. So since the original CNAME record has `_acme-challenge.` subdomain, the configured override domain must also have it.

If they just set it to `manage.clientsite.com`, then they would not need it in override domain, correct? And that is probably the normal use case.

So their need seems to be a bit of an edge case, in which specifying the _acme-challenge subdomain in the override domain is the correct solution.

I'll look into reverting my patch.

mholt:
Reverted in [9e63f36](https://github.com/caddyserver/certmagic/commit/9e63f36fa3e2b3260930e877ff014becafe73772)

crccw:
That's right, if the CNAME points to `manage.clientsite.com`, then override domain should be set to `manage.clientsite.com`.

PR #163, opened 2022-01-19
Managers: Ability to call GetCertificate from external certificate sources

mholt:
This is useful if another entity is managing certificates and can provide its own certs dynamically during handshakes.

For example, Tailscale's client has a `GetCertificate()` function: https://pkg.go.dev/tailscale.com/client/tailscale#GetCertificate

### DIFF
--- a/account.go
+++ b/account.go
@@ -62,7 +62,7 @@ func (am *ACMEManager) loadAccount(ca, email string) (acme.Account, error) {
 	if err != nil {
 		return acct, err
 	}
-	acct.PrivateKey, err = decodePrivateKey(keyBytes)
+	acct.PrivateKey, err = PEMDecodePrivateKey(keyBytes)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


mholt:
(I'm pretty much ready to merge this, but I am spending some time double-checking the On-Demand coupling to CustomGetCertificate, since I wonder if it's possible to use CustomGetCertificate without requiring on-demand.)

mholt:
And that's a wrap! I think. :slightly_smiling_face: 

Instead of `CustomGetCertificate` or `CertificateGetter`, I called the interface `CertificateManager`, since we expect that whatever is on the other end of `GetCertificate()` is managing the certificate for us (keeping it renewed).

It no longer implicitly enables or requires On-Demand TLS to be enabled; they are orthogonal concepts.

Thank you to Tailscale for making this change possible!

PR #166, opened 2022-01-31
Fix force-renewing revoked on-demand certs

mholt:
Follow-up to 9245be5a2f16135b9de55567682f49a2d343e1f0

Now I remember why I didn't implement this at first. Kind of complicated... and nobody at the time really needed it, as far as I knew.

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -472,7 +472,7 @@ func (cfg *Config) handshakeMaintenance(ctx context.Context, hello *tls.ClientHe
 
 	// Check OCSP staple validity
 	if cert.ocsp != nil && !freshOCSP(cert.ocsp) {
-		ocspResp, err := stapleOCSP(cfg.OCSP, cfg.Storage, &cert, nil)
+		err := stapleOCSP(cfg.OCSP, cfg.Storage, &cert, nil)
 		if err != nil {
 			// An error with OCSP stapling is not the end of the world, and in fact, is
 			// quite common considering not all certs have issuer URLs that support it.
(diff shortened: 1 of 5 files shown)


handshake.go
```diff
@@ -593,19 +604,23 @@ func (cfg *Config) renewDynamicCertificate(ctx context.Context, hello *tls.Clien
...
-			// even though the recursive nature of the dynamic cert loading
-			// would just call this function anyway, we do it here to
-			// make the replacement as atomic as possible.
-			newCert, err := cfg.CacheManagedCertificate(name)
-			if err != nil {
-				if log != nil {
-					log.Error("loading renewed certificate", zap.String("server_name", name), zap.Error(err))
+		if revoked {
+			return cfg.forceRenew(ctx, log, currentCert, revokedOCSP)
+		} else {
```
francislavoie, line 609: Since the `if` branch returns, I don't think you need the `else`, it can be unwrapped.

mholt: Thanks. I _knew_ something looked funny about that... been a long day.

mholt:
Ugh, this is so tricky.

Spent a couple days on this patch now and have only read through the code a dozen times to try to ensure it will be as correct as possible. Testing this functionality is so hard (last time it took me about a full day to set things up and test -- and even then, it's like one of those things where "the last time you fold up the parachute, you better do it right").

mholt:
I went ahead and implemented OCSP refreshing and revocation checking at load-time too (i.e. the call to `Manage()`), so that it won't take CertMagic up to an hour before scanning and discovering a revoked certificate if the staple is already not fresh.

This required some significant refactoring and it's super clear to me now why I didn't do this before.

Overall, this PR should make CertMagic's OCSP stapling and revocation checking more robust, and also including on-demand certificates now too, whereas it didn't before.

Very difficult to test this properly, so that will probably have to come later. (Feel free if anyone wants to help out!)

ocsp.go
```diff
@@ -93,33 +93,37 @@ func stapleOCSP(ocspConfig OCSPConfig, storage Storage, cert *Certificate, pemBu
...
 		gotNewOCSP = true
 	}
 
-	// By now, we should have a response. If good, staple it to
-	// the certificate. If the OCSP response was not loaded from
-	// storage, we persist it for next time.
+	if ocspResp.NextUpdate.After(cert.Leaf.NotAfter) {
+		// uh oh, this OCSP response expires AFTER the certificate does, that's kinda bogus.
+		// it was the reason a lot of Symantec-validated sites (not Caddy) went down
+		// in October 2017. https://twitter.com/mattiasgeniar/status/919432824708648961
```
francislavoie, line 104: This tweet got deleted it seems, unfortunately.

mholt:
As it turns out, this PR was almost correct. We later tested it in production and pushed a few more commits, and now we have good evidence that it works as expected.

PR #170, opened 2022-02-11
storage: Require fs.ErrNotExist (fix #168)

### LINKED ISSUE #168: Any error satisfies ErrNotExist
2f78e52

## What are you trying to do?

I am working on a module implementing the Storage interface. While testing my code, I noticed that any error satisfies the [ErrNotExist interface](https://github.com/caddyserver/certmagic/blob/master/storage.go#L280-L283). The following code is always true regardless of the error:

```
if _, ok := err.(certmagic.ErrNotExist); ok {
    // this is always true!
}
```

## What steps did you take?

Here is an example to illustrate the issue
https://go.dev/play/p/u-eYGshvKDp

`ErrNotExist` should not be an interface.

mholt:
Also stop using the deprecated io/ioutil package.
Update dependencies.
Update Go version in go.mod.

/cc @chaudyg

PS. Would this be a good time to propagate context into Storage, since external implementations will have to update anyway to use the new error semantics?


go.mod
```diff
@@ -1,13 +1,23 @@
 module github.com/caddyserver/certmagic
 
-go 1.14
+go 1.17
```
francislavoie, line 3: I don't think we can bump to 1.17 just yet, Fedora/RHEL don't ship with Go 1.17 yet. We might be able to bump it up in mid-May though.

hairyhenderson: `io/fs` was new in 1.16, so we should at least bump to 1.16

hairyhenderson: (though personally I'd advocate for going to 1.17 - I think being held back by RedHat is pointless 😝)

mholt: Does this line actually require anyone importing it to have that Go version or newer though?

In any case, I'm not super inclined to delay just a month or two for RedHat's benefit.

francislavoie: I'm pretty sure it does, yes.

francislavoie: Okay, but the implication is that we still can't use new features in this lib until Caddy is also ready to bump, else it won't compile.

I'm just not sure what the benefit is to marking the `go.mod` as 1.17 if we're not actually using 1.17 features here yet.

mholt: There are Go toolchain features offered by Go 1.17 that we can start using

chaudyg (approved):
thx @mholt for doing this. I would consider adding a test ! Appart from this, LGTM

chaudyg:
> PS. Would this be a good time to propagate context into Storage, since external implementations will have to update anyway to use the new error semantics?

I would say yes!

storage.go
```diff
@@ -35,7 +35,7 @@ import (
 // with the cluster.
 //
 // The Load, Delete, List, and Stat methods should return
```
chaudyg, line 37: What does `ErrNotExist` mean exactly for List? In the `filestorage` implementation, `ErrNotExist` is returned if the prefix is not a valid path. This seems a bit strange 🤔 .

mholt: Yeah; if you try to list the contents of a folder that doesn't exist, it should return ErrNotExist. Is that strange?

PR #171, opened 2022-02-15
Fixed order of certificate loading so that private keys are loaded first

sam-lord:
The order of storing the certificates was previously changed so that the private key would be stored first. For anyone who is creating storage hooks which push the certificate & key to a 3rd party service (like a CDN), the certificates are generally refused if uploaded before the private key.

Loading can trigger the same issue if (for any reason), the certificate & key have been deleted on the 3rd party service.

I'll admit that this use-case stretches what the storage system should be doing, but the change is trivial & just matches the storing side.


mholt (approved):
LGTM. I am a bit baffled as to why this matters at all, but I guess it doesn't hurt. :man_shrugging:

PR #175, opened 2022-02-23
increase acme client timeout to HTTPTimeout (30s)

Marmelatze:
With an internal ACME server (keyon) we had the problem that certificate requests always timed out. Increasing the timeouts from 15s to 30s solved the issue. The timeout does not seem to be configurable?

So if this has no side effects it could be raised to the global `HTTPTimeout` setting?


mholt (changes requested):
Hmmm,  are you still consistently getting slow connection times? 15+ seconds for regular TLS handshakes is absurdly long.

I think I'd rather make these separate variables that can be configured, instead of reusing `HTTPTimeout` for more than just HTTP.

mholt:
Since I didn't hear from you, I went ahead and just increased the timeouts in this commit: 55be6d8695034ccee7ada8a5314280131f3e51e3

PR #188, opened 2022-05-27
Set EDNS0 to 1232 bytes per recommendation

mohammed90:
See: https://caddy.community/t/hard-time-getting-a-response-on-a-dns-01-challenge/15721/16?u=mohammed90


mholt (approved):
Thanks Mohammed for the research. Let's give it a try

francislavoie:
Maybe we can get rid of this when we bump minimum to Go 1.19 btw (or do we get it for free by building on 1.19?)

mholt:
I'm guessing that since we're doing our own resolver, we won't get this patch for free just by using Go 1.19; we'd probably get it if we had Go do the lookups for us. But I'd have to read more into it to be sure.

mohammed90:
Matt is right. We're using github.com/miekg/dns here, not Go's DNS resolver. Miek's package does its own resolution and not rely on stdlib except for shuttling the net bytes.

PR #193, opened 2022-07-27
Use OverrideDomain when cleaning up DNS solver

gjung56:
Simple PR about OverrideDomain setting in DNS01Solver.
The Cleanup function isn't aware of the override so the txt record is never removed (as far I see)


mholt (approved):
Ah, thank you!

PR #199, opened 2022-08-11
Add one second (at most) to account for NotAfter imprecision

### LINKED ISSUE #197: incorrect time comparisons using the NotAfter timestamp

$ caddy version
v2.5.2 h1:eCJdLyEyAGzuQTa5Mh3gETnYWDClo1LjtQm2q9RNZrs=

## What are you trying to do?

Use certmagic with caddy to automatically renew certificates with certmagic as the ACME client.

## What steps did you take?

Here is my Caddyfile:

```
{
        debug
        acme_ca {$ACME_DIRECTORY_URL}
        acme_eab {
                key_id {$ACME_KID}
                mac_key {$ACME_HMAC_KEY}
        }
        admin off
        storage_clean_interval 10m
        renew_interval 1m
}
{$DNS_NAME} {
  respond "Hello There!"
}
```

## What did you expect to happen, and what actually happened instead?

I expected certmagic to track a certificates validity until the end of its life, but certmagic drops the last (fractional) second from the certificate life due to the imprecise definition and granularity of `Not After` in X.509 certificates. [More background and discussion on this issue here](https://community.letsencrypt.org/t/discussion-of-2021-06-08-certificate-lifetime-incident-valid-for-an-extra-one-second/153469). The tl;dr is that the certificate is valid for the entire second of the `Not After` value, so a certificate with a `Not After` value of `2022/01/01 23:59:59` is still valid at `2022/01/01 23:59:59.999`.

This is causing the following log message to appear in the caddy logs, because the certificate's `Not After` is compared to the OCSP's `nextUpdate` (which does not have the same imprecision issue), without accounting for the different granularities.

```
tls.cache.maintenance   stapling OCSP   {"identifiers": ["foo.sample.internal"], "error": "invalid: OCSP response for [foo.sample.internal] valid after certificate expiration (-1s)"}
```

certmagic should be aware of this issue when doing time comparisons, for example in the OCSP check and cache expiration timestamp.

I'm happy to open a PR to fix the time comparison issue. If so, would you prefer to have a `Certificate.ValidBefore time.Time` field with the timestamp that can be used for comparison, or update all comparisons to account for the extra `(1 * time.Second)`?

## Please link to any related issues, pull requests, and/or discussion

https://community.letsencrypt.org/t/discussion-of-2021-06-08-certificate-lifetime-incident-valid-for-an-extra-one-second/153469

## Bonus: What do you use CertMagic for, and do you find it useful?

Testing out automated certificate rotation with caddy/certmagic.

benburkert:
Here is a PR to fix #197. I ended up going with a function instead of a method on `Certificate` because there are a couple places where the `NotAfter` of a `x509.Certificate` are being used. I also had a hard time figuring out how to write a test for this. It looked like the filesystem storage cache needs to be populated with a certificate and ocsp response, but i didn't see a straightforward way or any prior examples of this in other tests. Let me know if you have any thoughts on how to test this.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -197,7 +197,7 @@ func (certCache *Cache) unsyncedCacheCertificate(cert Certificate) {
 		if certCache.logger != nil {
 			certCache.logger.Debug("certificate already cached",
 				zap.Strings("subjects", cert.Names),
-				zap.Time("expiration", cert.Leaf.NotAfter),
+				zap.Time("expiration", expiresAt(cert.Leaf)),
 				zap.Bool("managed", cert.managed),
 				zap.String("issuer_key", cert.issuerKey),
 				zap.String("hash", cert.hash))
@@ -242,7 +242,7 @@ func (certCache *Cache) unsyncedCacheCertificate(cert Certificate) {
 	if certCache.logger != nil {
 		certCache.logger.Debug("added certificate to cache",
 			zap.Strings("subjects", cert.Names),
-			zap.Time("expiration", cert.Leaf.NotAfter),
+			zap.Time("expiration", expiresAt(cert.Leaf)),
 			zap.Bool("managed", cert.managed),
 			zap.String("issuer_key", cert.issuerKey),
 			zap.String("hash", cert.hash),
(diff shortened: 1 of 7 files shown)


mholt (approved):
Thank you for this patch. It looks clean and makes sense!

PR #212, opened 2022-12-08
Allow specifying http proxy via config

georgmu:
Currently, it is only possible to specify the http proxy via environment variable.

For a more flexible setup, it is required to set the proxy function directly.

This pull request adds a new field HttpProxy to the config (with the default value http.ProxyFromEnvironment), allowing to overwrite the proxy or to disable it completely.

### DIFF
--- a/acmeissuer.go
+++ b/acmeissuer.go
@@ -115,6 +115,10 @@ type ACMEIssuer struct {
 	// desired, set this to zap.NewNop().
 	Logger *zap.Logger
 
+	// Set a http proxy to use when issuing a certificate.
+	// Default is http.ProxyFromEnvironment
+	HTTPProxy func(*http.Request) (*url.URL, error)
+
 	config     *Config
 	httpClient *http.Client
 
@@ -204,6 +208,13 @@ func NewACMEIssuer(cfg *Config, template ACMEIssuer) *ACMEIssuer {
 		template.Logger = defaultLogger
 	}
 
+	if template.HTTPProxy == nil {
+		template.HTTPProxy = DefaultACME.HTTPProxy
+	}
+	if template.HTTPProxy == nil {
+		template.HTTPProxy = http.ProxyFromEnvironment
+	}
+
 	template.config = cfg
 	template.mu = new(sync.Mutex)
 
@@ -223,7 +234,7 @@ func NewACMEIssuer(cfg *Config, template ACMEIssuer) *ACMEIssuer {
 		}
 	}
 	transport := &http.Transport{
-		Proxy:                 http.ProxyFromEnvironment,
+		Proxy:                 template.HTTPProxy,
 		DialContext:           dialer.DialContext,
 		TLSHandshakeTimeout:   30 * time.Second, // increase to 30s requested in #175
 		ResponseHeaderTimeout: 30 * time.Second, // increase to 30s requested in #175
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -123,6 +124,10 @@ type Config struct {
 	// a default logger will be created.
 	Logger *zap.Logger
 
+	// Set a http proxy to use when issuing a certificate.
+	// Default is http.ProxyFromEnvironment
+	HttpProxy func(*http.Request) (*url.URL, error)
```
mholt, line 129:
```suggestion
	HTTPProxy func(*http.Request) (*url.URL, error)
```

In this house, we capitalize initialisms :upside_down_face: 

Also, I think this would be better placed in the ACMEIssuer struct, as it's used only for ACME transactions.

mholt (changes requested):
Thanks for the enhancement!

This is acceptable, but I think it should be configured directly on the ACMEIssuer rather than the CertMagic config.

config.go
georgmu: Ok, I will change both.
When in ACMEIssuer, I can either default to `http.ProxyFromEnvironment` (to be backwards-compatible), but this way there is no option to disable the proxy.
The other option would be to default to `nil`, which allows to explicitly disable the proxy for ACME requests, but slightly changes backwards compatibility.

georgmu:
The updated commit defaults to `nil` proxy.

I could add the following:
```
if template.HTTPProxy == nil {
    template.HTTPProxy = http.ProxyFromEnvironment
}
```
This would forbid to disable the proxy entirely (which is okay for me), but make it backwards-compatible with the current behavior.

georgmu:
I changed my mind and implemented it backwards-compatible.
If someone wants to completely disable proxy, one could pass the following:
```
template.HTTPProxy = func(r *http.Request) (*url.URL, error) { return nil, nil }
```

georgmu:
There are still some requests not using the proxy. I will see if I can find the right place

georgmu:
Please ignore my last comment. There was an error in my test setup.

From my side, the tests look good regarding the current pull request.

acmeissuer.go
```diff
@@ -198,6 +202,9 @@ func NewACMEIssuer(cfg *Config, template ACMEIssuer) *ACMEIssuer {
 	if template.Logger == nil {
 		template.Logger = DefaultACME.Logger
 	}
+	if template.HTTPProxy == nil {
+		template.HTTPProxy = http.ProxyFromEnvironment
```
mholt, line 206: We should get this default value from `DefaultACME`, otherwise if the user sets `DefaultACME.HTTPProxy`, it will have no effect.

```suggestion
		template.HTTPProxy = DefaultACME.HTTPProxy
```

acmeissuer.go
```diff
@@ -223,7 +230,7 @@ func NewACMEIssuer(cfg *Config, template ACMEIssuer) *ACMEIssuer {
 		}
 	}
 	transport := &http.Transport{
-		Proxy:                 http.ProxyFromEnvironment,
+		Proxy:                 template.HTTPProxy,
```
mholt, line 233: We can apply the default of `http.ProxyFromEnvironment` (around) here instead, if `template.HTTPProxy` is nil.

mholt (changes requested):
Thanks for the updates @georgmu! Just one more small suggestion, if that's ok. There's the "user-defined default" that we need to apply first, from DefaultACME, then when going to actually make the transport, we can use `http.ProxyFromEnvironment` if no default is set.

georgmu:
Thanks, @mholt.
I implemented the suggested changes.

Personally, I would skip the second nil check: if someone intentionally sets DefaultACME.HTTPProxy to nil, the default (ProxyFromEnvironment) will be used anyway, which might be unknowingly.
And - in contrast to the logger which gets handled again - it is not really necessary to have a proxy.

But both is fine for me.

mholt (approved):
Ok, that's a fair point. Let's try this then. Thanks for the contribution!

PR #213, opened 2023-01-10
Allow the default cache logger to be set

pwilloughby:


mholt (approved):
Thanks! I think this is probably sensible. Have you noticed this benefit you? How?

pwilloughby:
> Have you noticed this benefit you? How?

This benefits me because it allows changing the cache logger when using the high-level convenience functions HTTPS(), Listen(), TLS(). I think the only way to set the cache logger currently is to make your own cache.

mholt:
Yeah, that's true. Thanks for the contribution!

PR #225, opened 2023-04-11
Add a `GetCertificateWithContext` function

ankon:
This basically would allow us to pass a context into the "certmagic world", from which it then would get passed through back to our storage implementation.

The reason we want that is so that we can actually set up application performance monitoring by instrumenting the calls.

Note that this does add a new public function, but unless you specifically set your `GetCertificate` in the TLSConfig to something that uses this function, everything should stay the same.

Our setup looks like this, very simplified:

```golang
certmagicConfig.Issuers = []certmagic.Issuer{acmeIssuer}

tlsConfig := certmagicConfig.TLSConfig()
	txn := monitoring.StartTransaction("TLS/GetCertificate")
	defer txn.End()
	txn.AddAttribute("supportedVersions", supportedVersionsToString(clientHelloInfo.SupportedVersions))

	// ... do some checks here to make sure we want to handle this request

	ctx := newrelic.NewContext(context.Background(), txn)
	if err != nil {
		// ... review the error and possible ignore it/handle it differently
	}
	return cert, err
}
```

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -42,9 +42,15 @@ import (
 // 5. Issuers (if on-demand is enabled)
 //
 // This method is safe for use as a tls.Config.GetCertificate callback.
+//
+// GetCertificate will run in a new context, use GetCertificateWithContext to provide
+// a context.
 func (cfg *Config) GetCertificate(clientHello *tls.ClientHelloInfo) (*tls.Certificate, error) {
 	ctx := context.TODO() // TODO: get a proper context? from somewhere...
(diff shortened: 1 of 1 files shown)


mholt:
Hi, sorry for the lack of reply -- I'll be getting around to this soon as I prepare the Caddy 2.7 beta 1 release :)

mholt (approved):
LGTM. Thank you for the enhancement!

PR #227, opened 2023-04-16
Update & Improve CI

wusatosi:
Hi, this CI updates the CI workflow file.
This PR:
- Updates `checkout` v2 -> v3
- Updates `setup-go` v2 -> v4
- Configures `setup-go` to use the go version consistent with project setup via `go.mod` file
- Delegates build caching to `setup-go`
- Changes dependency installation to `go mod download` ([official documentation](https://go.dev/ref/mod#go-mod-download), [difference with the current `go get`](https://stackoverflow.com/questions/66356034/what-is-the-difference-between-go-get-command-and-go-mod-download-command))
- Replaces the deprecated `set-output`
- Uploads coverage file as an artifact, as currently the generated file is discarded

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,5 +50,5 @@ jobs:
     - name: Archive code coverage results
       uses: actions/upload-artifact@v3
       with:
-        name: code-coverage-report
+        name: ${{ runner.os }}-code-coverage-report
         path: "cover-profile.out"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,28 +39,11 @@ jobs:
         env
         # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
-
-#   - name: Cache the build cache
-#     uses: actions/cache@v2
-#     with:
-#       # In order:
-#       # * Module download cache
-#       # * Build cache (Linux)
-#       # * Build cache (Mac)
-#       # * Build cache (Windows)
-#       path: |
-#         ~/go/pkg/mod
-#         ~/.cache/go-build
-#         ~/Library/Caches/go-build
-#         ~\AppData\Local\go-build
-#       key: ${{ runner.os }}-${{ matrix.go }}-go-ci-${{ hashFiles('**/go.sum') }}
-#       restore-keys: |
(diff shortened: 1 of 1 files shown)


wusatosi:
Btw, what's the use for `Print Go version and environment`? `go env` is displayed via `setup-go` step, printing environmental variables is not safe as it may contain secrets?

francislavoie:
Thanks for this!

We print that stuff out mostly to aid in debugging. When playing with the CI config, it usually saves time to have immediate access to the envs as a reference.

`go env` didn't used to be printed I think. And I don't think there's any relevant secrets in the system env vars in this case (secrets typically get fed through CI variables which aren't in env).

Most of the things that look pointless in here are largely because we copy-pasted from https://github.com/caddyserver/caddy/blob/master/.github/workflows/ci.yml which has some additional steps.

Like `short_sha` isn't actually used in here, it was used in the `caddy` repo to attach the commit sha to the binary artifacts. This project is a lib, so there's no binary to publish.

I think we should probably be testing on latest Go + minimum from `go.mod`, so I'm not sure using `go-version-file: 'go.mod'` is ideal since that approach isn't compatible with a matrix.

I'm not sure publishing the coverage report on its own is that useful, we're not using any tooling to analyze it currently. I think Matt's opinion is that code coverage is usually a misleading metric, so it's not of great importance to us.

wusatosi:
Oh, I just saw your PR in caddy's CI. Lucky we both found this out at (almost the same time).

Cool, I guess I will remove the `short_sha` thing, revert the `go-version-file`, and disable coverage reporting.

wusatosi:
I don't think there's a minimum compatible go version declared in docs. I will just leave go version matrix in CI as is.
@francislavoie should I add  1.19, 1.20 as is in Caddy?

francislavoie:
The minimum is what's in `go.mod` basically. I think we can test on `1.18` + `1.20` for now.

wusatosi:
> The minimum is what's in `go.mod` basically. I think we can test on `1.18` + `1.20` for now.

Updated

francislavoie:
LGTM! I don't actually have merge permissions in this repo 😅

wusatosi:
> LGTM! I don't actually have merge permissions in this repo 😅

ops haha no worries

mholt (approved):
Silly me. Thanks both :)

PR #229, opened 2023-04-16
chore: Skip slow tests on Windows

### LINKED ISSUE #228: Windows test taking significantly longer time than macos/ ubuntu
Currently, in CI testing. 
The windows testing pipeline takes the majority amount of time.
image

According to the log.
Looks like `TestFindZoneByFqdn` is spending 160s on [windows](https://github.com/caddyserver/certmagic/actions/runs/4670033562/jobs/8269270484#step:7:82).
image

While the same test spends under a second on [Ubuntu](https://github.com/caddyserver/certmagic/actions/runs/4670033562/jobs/8269270198#step:7:82)/ [Macos](https://github.com/caddyserver/certmagic/actions/runs/4670033562/jobs/8269270344#step:7:82).

image

Relevant code here:
https://github.com/caddyserver/certmagic/blob/462f5e2334a762adc04d2e9ca6200239ffb49a5c/dnsutil_test.go#L87-L177

francislavoie:
Fixes #228

Essentially, Windows takes a very long time to give up with bad resolver addresses, waiting the maximum amount of time for an `i/o timeout`.

We can just ignore these tests on Windows. It works, it's just really slow.

### DIFF
--- a/dnsutil_test.go
+++ b/dnsutil_test.go
@@ -9,6 +9,7 @@ package certmagic
 import (
 	"net"
 	"reflect"
+	"runtime"
 	"sort"
 	"strings"
 	"testing"
@@ -91,6 +92,7 @@ var findXByFqdnTestCases = []struct {
 	primaryNs     string
 	nameservers   []string
 	expectedError string
+	skipTest      bool
 }{
 	{
 		desc:        "domain is a CNAME",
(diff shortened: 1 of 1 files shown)


mholt (approved):
I guess this is as good a solution as any. The other OSes should be able to prove it out for us. Thanks.

(Is this something we should enhance in the code so that things don't take as long on Windows?)

francislavoie:
I don't think there's any information we can use to know if it should fail quicker. I played with the timeout options on the DNS client and all it did is shorten the maximum amount of time before it fails.

Btw the default timeouts are 2s on the DNS client, I'm not sure we need it to be 10s, not sure what the usecase was for making it that long, but at this point I don't want to risk breaking anyone.

mholt:
I would be surprised if anyone is _relying_ on the timeout being a full 10s... :thinking: I'm not sure who would do that :smile:

francislavoie:
Yeah I mean I'm just being careful/paranoid :sweat_smile:

Like really, it's not a problem that it takes long to fail under normal circumstances. The problem is only relevant specifically during testing because we try it so often and multiple times per test run.

If we ever make changes to this part of the code, we can just comment out the skip to run it in CI one time during a PR, then turn it back off before merging. It's a waste of time to test it every time, obviously.

Issue #244, opened 2023-07-13
Feature request: exported variable for the "no OCSP server specified in certificate" error


kenjenkins:
## What would you like to have changed?

Could certmagic define an exported error value for the "no OCSP server specified in certificate" error returned here? https://github.com/caddyserver/certmagic/blob/d8b13df4d123be9737c6855b78fe3417df8bad11/ocsp.go#L152

For example:
```go
var ErrNoOCSPServerSpecified = errors.New("no OCSP server specified in certificate")
```

## Why is this feature a useful, necessary, and/or important addition to this project?

When using a mix of certificates (some managed by certmagic, and some not), we can end up with some spurious "no OCSP stapling for _[domain]_" errors, for those certificates not managed by certmagic. I'd like to selectively ignore this specific error, while still logging any other errors from the OCSP code.

If the certmagic API included an exported error value for this case, then we could use `errors.Is()` to check for this specific error. This feels more "correct" to me than matching against the error message text.

I expect the maintenance burden for the exported error value to be minimal.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

In the meantime, we can do a `strings.Contains()` check on the error message, looking for the substring "no OCSP server specified in certificate".

## Please link to any relevant issues, pull requests, or other discussions.

n/a

francislavoie:
Seems reasonable. You could open a PR!

kenjenkins:
Thanks @francislavoie, I've opened a PR. I skimmed the [contributing guidelines](https://github.com/caddyserver/certmagic/blob/master/.github/CONTRIBUTING.md), but please let me know if there are any steps I've missed.

mholt:
Thank you for the request and the contribution!

PR #245, opened 2023-07-13
Define a NoOCSPServerSpecified error

kenjenkins:
Define an exported NoOCSPServerSpecified error, to make it easier to distinguish the case that a certificate does not support OCSP (from other OCSP stapling errors). Add a unit test exercising this behavior.

Implements #244.

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/pem"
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thanks for the PR! This is a high-quality and simple contribution, which we welcome. :+1: Happy to merge this as soon as CI completes.

PR #248, opened 2023-08-08
Create a new context when renewing a certificate in the background

ankon:
The context available to `renewDynamicCertificate` comes from inside the TLS handshake, and as such may be bounded by the lifespan of the connection. Passing this into a goroutine will lead to problems when the connection ends (and the connection context gets canceled with it) but the goroutine is going to do more I/O on that context.

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -727,7 +727,7 @@ func (cfg *Config) renewDynamicCertificate(ctx context.Context, hello *tls.Clien
 
 	// if the certificate hasn't expired, we can serve what we have and renew in the background
 	if timeLeft > 0 {
-		ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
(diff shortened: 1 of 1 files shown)


ankon:
FWIW, `go test -v -short -race ./...` fails locally:

```
=== RUN   TestFindZoneByFqdn/domain_is_a_eTLD
    dnsutil_test.go:172: test 2: expected no error, but got: could not find the start of authority for example.com.ac.: NXDOMAIN
    dnsutil_test.go:175: test 2: expected zone '' but got 'ac.'
```

mholt (approved):
Just seeing this -- exactly the change I was going to make after I read the issue.

Thanks for the patch!

mholt:
> FWIW, go test -v -short -race ./... fails locally:

Yeah, the CertMagic tests currently use live DNS which I kind of hate :sweat_smile: I just haven't done the work to mock them up yet. They can be problematic depending on which environments you run them in.

Issue #247, opened 2023-08-08
Handling canceled contexts in `Unlock` in custom storage implementations


ankon:
## What is your question?

This relates to this piece of code: 
https://github.com/caddyserver/certmagic/blob/51b31907aaa07f21e1bf2974193d11cdbf32802b/config.go#L504-L517

Basically: It is possible that locking works, but the context gets subsequently canceled for whatever reason. The deferred `releaseLock` call in line 511 will call our custom storage implementation's `Unlock` with the now-canceled context. Right now that leads to an error in our code, and eventually our lock will expire and we can continue.

Is there anything we can do better here? For instance, would it be reasonable to pass another context into `releaseLock`/`Unlock` that is not canceled? Or should we "ignore" the cancelation on our side?

For reference: Our storage uses DynamoDB, and `Lock` and `Unlock` basically modify a specific item in it. We pass the context through to the AWS SDK DynamoDB call, which will recognize it is canceled and not execute the actual request.

mholt:
> Is there anything we can do better here? For instance, would it be reasonable to pass another context into releaseLock/Unlock that is not canceled? Or should we "ignore" the cancelation on our side?

Good question! Hmm.

Part of me thinks, ignore the context because the point is to cleanup after the context is cancelled. The other part of me thinks, if the context is cancelled, it's important to abort to avoid a goroutine leak (or something).

Or maybe we SHOULD make a *new* context that isn't canceled, but has a timeout, so that unlock can still finish, but the time is set. Is that reasonable? Or do you think the storage implementation should have control of the timeout and set its own when it encounters an already-canceled context?

> I found the underlying cause of the canceled context by now: The background renewal of a not-yet-expired certificate is using the same context as was given to GetCertificateWithContext, which in our case is actually the handshake context coming from https://pkg.go.dev/crypto/tls#ClientHelloInfo.Context.
>
> See also https://github.com/caddyserver/certmagic/pull/248: The background renewal should use a different context, IMHO it is wrong to assume that it can use the one given to GetCertificate, which is supposed to be called inside the TLS handshake and as such does have a limited time to do things.

I think I agree. Hadn't really considered this, but a background routine indeed should not be confined to a single handshake.

But why haven't we encountered this problem before? :thinking:

ankon:
> Or maybe we SHOULD make a new context that isn't canceled, but has a timeout, so that unlock can still finish, but the time is set.

[...]

Issue #246, opened 2023-08-07
Renewal op retries when an expired certificate cannot be renewed


ankon:
## What version of the package are you using?

github.com/caddyserver/certmagic v0.19.1

## What are you trying to do?

We're using certmagic to manage certificates for a number of hostnames we host websites for. Unfortunately we do not control DNS for these websites, and it can happen that users accidentally break parts of the setup that prevent certificate renewals. 

## What steps did you take?

This is somewhat theoretical, but based on our current understanding of what we're seeing:

1. Have a expired certificate in storage (and possibly cached?)
2. Prevent ACME challenges from succeeding (i.e. no HTTP-01, no TLS-ALPN-01, no DNS-01 should work)
3. Request the certificate (in our case: This happens through the `GetCertificate` function as part of the handshake)

## What did you expect to happen, and what actually happened instead?

I would expect an error: The certificate is expired, and shouldn't be served, nor is it in its "renewal window". 
I would also expect that the error will take a while to be returned, because presumably we would see a request to obtain a new certificate.

What actually seems to happen is that certmagic perpetually tries to renew the certificate, times out (expected), and then retries renewal.

From our monitoring system a stack trace, with the operation at this point having run for almost 14h
![image](https://github.com/caddyserver/certmagic/assets/1210641/cf313310-5b1c-4285-b3fa-526753205753)

## How do you think this should be fixed?

Couple of things I found interesting and possibly relevant here:

1. An error from `handshakeMaintenance` is logged but ignored in `loadCertFromStorage`. This might be ok and intentional, but I wonder whether it should check the type of error ("is this ok to defer" vs "does this block any usage of the certificate?"): https://github.com/caddyserver/certmagic/blob/51b31907aaa07f21e1bf2974193d11cdbf32802b/handshake.go#L405-L410
2. `currentlyInRenewalWindow` returns `true` even when the certificate is expired: https://github.com/caddyserver/certmagic/blob/51b31907aaa07f21e1bf2974193d11cdbf32802b/certificates.go#L88-L103 . I would expect it to return `false` in that case, and the caller to explicitly do an obtain-the-certificate.

A bit as an aside: The timeouts involved in renewing ("2 minutes", "90s" when expired, "5min" when time left) look a bit arbitrary, should these be made configurable? 

## Bonus: What do you use CertMagic for, and do you find it useful?

We absolute do find it useful! :)

mholt:
> We absolute do find it useful! :)

[...]

PR #249, opened 2023-08-08
Use context from ClientHello during GetCertificate

mholt:
See discussion in #247

/cc @ankon -- I haven't tested this myself yet (been a very, very busy day with notifications rolling in all day!) but I feel like this makes the most sense. Could you try it when you have a chance?

Thanks for your patience!

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -274,7 +274,7 @@ func (cfg *Config) getCertDuringHandshake(ctx context.Context, hello *tls.Client
 	}
 
 	// By this point, we need to load or obtain a certificate. If a swarm of requests comes in for the same
-	// domain, avoid pounding manager or storage thousands of times simultaneously. We do a similar sync
+	// domain, avoid pounding manager or storage thousands of times simultaneously. We use a similar sync
 	// strategy for obtaining certificate during handshake.
 	certLoadWaitChansMu.Lock()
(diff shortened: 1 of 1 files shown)


ankon:
So, we actually were running with this for a while, but reverted it yesterday due to the problems in #247/#248. Apart from that this worked quite like it should though, so I _think_ this would be good to do indeed.

I'm right now not sure if #246 plays into this as well, gut feeling says "maybe, but probably not" as it is a synchronous case and the handshake context itself doesn't have a timeout (but a I/O deadline on the underlying network connection!).

mholt:
Gotcha; ok I'll hold this open until we are confident with a fix for #246 and can try them together.

ankon:
> Could you find a safe way to try it out and let me know?

We're currently redoing our infra to run experiments like this in a safer way, so will try it out as soon as we that up and running :)

mholt:
Awesome. In the future, as we work on testing infrastructure with the Caddy project, I'm hoping this kind of thing will get easier :)

mholt:
I've been using this a little bit and so far it is working well. I want to get this fix, along with a patch I just merged into this branch, out for Caddy v2.7.4 today, so I'm going to merge this in and tag CertMagic v0.19.2.

When you do get around to trying this out, just try the latest release (or commits on master). Looking forward to your results! :100:

PR #270, opened 2024-02-19
Demote "storage cleaning happened too recently" from WARN to INFO

francislavoie:
This isn't an actionable problem, so it doesn't make sense as a WARN.

It shows up in startup logs pretty often if starting and stopping Caddy, slightly annoying.


mholt (approved):
Yeah, thanks. This is better.

PR #271, opened 2024-02-25
inject http01solver

2456868764:


2456868764:
LGTM

mholt:
What is this for? How about a little explanation?

2456868764:
> What is this for? How about a little explanation?

For example, there is a scenario where there is a certificate management service that uses the certmagic library and is deployed behind the gateway. When authenticates through http-01 solver, it not only needs to start the web service to accept the authentication request, but also needs to create an Ingress route in the gateway which forwards the authentication request to the web service, so a custom HttpSolver needs to be injected to override the default implementation.

mholt:
Interesting. I think most cases like this, you have the solver sit in front of CertMagic and/or its HTTP handler. Is that something that would work for you too?

mholt:
Could this be done by injecting just the HTTP handler instead of replacing the whole solver? I might be exploring this.

mholt:
@2456868764 I've pushed a change here that allows you to set your own Handler in the HTTP-01 solver, which is effectively the ability to inject the solver itself:

https://github.com/caddyserver/certmagic/pull/279/files#diff-0d423173eaa496408ee9e7d82ae770feaa3b860783ca6a155f6d7551cb5d4d14R50

I will be merging that soon, superseding this, if that's alright. Let me know if that doesn't work for you though! Thanks for the submission!

Issue #219, opened 2023-03-15
Issue with resolvers


MisterDuval:
Hi,
I'm having issue with resolvers, my caddyFile looks like:

[...]
tls {
   dns cloudflare MY_API_KEY
   resolvers 1.1.1.1
}
But I'm getting this error because IT blocks DNS requests except from 1.1.1.1.
"error": "[URL] solving challenges: waiting for solver certmagic.solverWrapper to be ready: checking DNS propagation of "_acme-challenge.URL": dial tcp 108.162.195.175:53: i/o timeout

This IP is related to Cloudflare NS

mholt:
Did you verify that the config you think you're running is actually the config that's running?

In isolation, the config you gave works. What is your full config that we can reproduce the issue? (Including domain names)

MisterDuval:
Yes, my config file is valid and tested, maybe that "resolvers" is just ignored? 
As soon as I allowed that other DNS IP (108.162.195.175) in firewall, it has worked.  You can try with a `ca https://acme-staging-v02.api.letsencrypt.org/directory` and a firewall blocking all DNS requests except 1.1.1.1 (port TCP 53)

Caddy v2.6.4 (on Windows), compiled with xcaddy and caddy-exec + caddy-dns-cloudflare

caddyFile:
```
{
    email MY_EMAIL
}

https://mycustomdomain.ltd
{
    tls {
   	    dns cloudflare MYAPIKEY
	    resolvers 1.1.1.1
	}

    root htdocs

    encode gzip
    file_server
    php_fastcgi 127.0.0.1:9123
    log {
       output file logs/caddy-error.log
       level ERROR
    }
}
```

mholt:
Thanks for the follow-up. It seems unlikely that they're being ignored, however, if you have a moment to try putting a `fmt.Println()` line just before this line:

https://github.com/caddyserver/certmagic/blob/3514797a52dd6e6ac316eeb791545885ffbc33b0/solvers.go#L375

And maybe a few more in the call stack down the `checkDNSPropagation()` function would likely prove which resolvers are being used. If it's not 1.1.1.1 then it should be easier to figure out why.

mholt:
@MisterDuval Any interest in digging into this?

MisterDuval:
Sorry I had no time for this for now.

mholt:
No worries. I'll probably close it for now, since I think we'll need your help to troubleshoot this. We can reopen when ready :+1:

[...]

PR #274, opened 2024-03-12
Check DNS propagation at authoritative nameservers only with default resolvers

pgeh:
Discussion in https://github.com/caddyserver/certmagic/issues/219


mholt (approved):
Let's give this a shot. Hopefully it works well for everyone too! :) Thank you!

PR #277, opened 2024-03-20
Update readme examples to use TLS-ALPN const from ACMEz

goksan:
There's a couple of references to `tlsalpn01.ACMETLS1Protocol` in the readme. I've changed them to `acmez.ACMETLS1Protocol`.

I think these should have been changed when CertMagic switched from lego to ACMEz: https://github.com/caddyserver/certmagic/commit/b76b76abfcacf44350226e59907fbdc23b31f489


mholt (approved):
Aha, thank you! I did miss those.

PR #283, opened 2024-04-18
Use the `email` configuration in the ACME issuer to "pin" an account to a key

ankon:
When the issuer is configured with both an email and key material, these should match -- but that also means we can use the email information to predict the key-key, skipping the potentially expensive storage.List operation.

### DIFF
--- a/account.go
+++ b/account.go
@@ -88,11 +88,19 @@ func (*ACMEIssuer) newAccount(email string) (acme.Account, error) {
 // If it does not exist in storage, it will be retrieved from the ACME server and added to storage.
 // The account must already exist; it does not create a new account.
 func (am *ACMEIssuer) GetAccount(ctx context.Context, privateKeyPEM []byte) (acme.Account, error) {
-	account, err := am.loadAccountByKey(ctx, privateKeyPEM)
-	if errors.Is(err, fs.ErrNotExist) {
-		account, err = am.lookUpAccount(ctx, privateKeyPEM)
+	email := am.getEmail()
+	if email == "" {
+		account, err := am.loadAccountByKey(ctx, privateKeyPEM)
+		if err == nil || !errors.Is(err, fs.ErrNotExist) {
+			return account, err
+		}
+	} else {
+		keyBytes, err := am.config.Storage.Load(ctx, am.storageKeyUserPrivateKey(am.CA, email))
+		if err == nil && bytes.Equal(bytes.TrimSpace(keyBytes), bytes.TrimSpace(privateKeyPEM)) {
+			return am.loadAccount(ctx, am.CA, email)
+		}
 	}
-	return account, err
+	return am.lookUpAccount(ctx, privateKeyPEM)
 }
 
 // loadAccountByKey loads the account with the given private key from storage, if it exists.
--- a/account.go
+++ b/account.go
@@ -90,9 +90,8 @@ func (*ACMEIssuer) newAccount(email string) (acme.Account, error) {
 func (am *ACMEIssuer) GetAccount(ctx context.Context, privateKeyPEM []byte) (acme.Account, error) {
 	email := am.getEmail()
 	if email == "" {
-		account, err := am.loadAccountByKey(ctx, privateKeyPEM)
-		if err == nil || !errors.Is(err, fs.ErrNotExist) {
-			return account, err
+		if account, err := am.loadAccountByKey(ctx, privateKeyPEM); err == nil {
+			return account, nil
 		}
 	} else {
 		keyBytes, err := am.config.Storage.Load(ctx, am.storageKeyUserPrivateKey(am.CA, email))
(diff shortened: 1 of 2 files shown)


account.go
```diff
@@ -88,11 +88,19 @@ func (*ACMEIssuer) newAccount(email string) (acme.Account, error) {
 // If it does not exist in storage, it will be retrieved from the ACME server and added to storage.
 // The account must already exist; it does not create a new account.
 func (am *ACMEIssuer) GetAccount(ctx context.Context, privateKeyPEM []byte) (acme.Account, error) {
-	account, err := am.loadAccountByKey(ctx, privateKeyPEM)
-	if errors.Is(err, fs.ErrNotExist) {
-		account, err = am.lookUpAccount(ctx, privateKeyPEM)
+	email := am.getEmail()
+	if email == "" {
+		account, err := am.loadAccountByKey(ctx, privateKeyPEM)
+		if err == nil || !errors.Is(err, fs.ErrNotExist) {
```
ankon, line 94: `loadAccountByKey` can return different error types: "whatever storage returned for a non-existing key" and `fs.ErrNotExist` when there is indeed no account at all.

There doesn't seem to be a requirement for the storage to return fs error codes, and it also seems a bit odd that it aborts the first time and not `continue`s to the end of the list of accounts:
https://github.com/caddyserver/certmagic/blob/f64401c80d50d396f1cedefb316108c10ebcbf84/account.go#L108-L111

WDYT: Would a `continue` there make more sense, and then in `GetAccount` we can just drop the `!errors.Is(err, fs.ErrNotExist)` half and always look up the account if one wasn't found by `loadAccountByKey`?

ankon:
Mostly for tracking: This PR is against master, but we're actually still using 0.20.0 (no ACME v2). I have a separate branch for that at https://github.com/framer/certmagic/tree/use-account-email-to-pin_0.20.0, the code changes are identical.

account.go
mholt: > loadAccountByKey can return different error types: "whatever storage returned for a non-existing key" and fs.ErrNotExist when there is indeed no account at all.

Hmm, I would expect the storage implementation to return `fs.ErrNotExist` whenever the key or account does not exist. If the key doesn't exist, that means the account doesn't exist (in storage). Thus it needs to be looked up. (Maybe I'm missing something though)

> There doesn't seem to be a requirement for the storage to return fs error codes,

Maybe I should clarify that in the docs. My understanding though is that `fs` errors are conventional now, and Go std lib uses and recommends them all over the place (os, io packages for example). We can clarify that in the storage API godocs if needed. :+1:

> WDYT: Would a continue there make more sense, and then in GetAccount we can just drop the !errors.Is(err, fs.ErrNotExist) half and always look up the account if one wasn't found by loadAccountByKey?

Yes I suppose that would make sense, but I think an error there implies a broken storage backend or a corrupted file structure (i.e. a file that does not exist but should). So I dunno... MAYBE ignoring is the right thing to do there... ?

ankon: If the storage is corrupted indeed it could be a different problem, so continuing could also be stupid. Or you're lucky, and the thing that's corrupted wasn't what you were looking for anyways, and the thing you _are_ looking for is the next on the list.

The storage listing said there would be a prefix for that email, but that doesn't say anything about the files inside that prefix. If something failed to write the key file into that prefix earlier, we would now just fail -- regardless whether this is the right thing in the first place. IOW: The storage corruption in this case isn't "disk is failing" but "_we_ earlier failed". Then again, if you do fast-exit here, we go into "lookup via key from ACME", and assuming that key was correct then we will now retry to save the account -- regardless whether that was indeed this broken key (possibly healing the earlier problem) or another already existing key (which we could have found, had we not stopped). 

`continue`-ing here avoids the second case problem: If what is broken is not what we're looking for, and what we're looking for does exist, we will find it, and avoid the ACME call + redundant save.

mholt: I'm convinced about the continue. That's probably right. Thanks for hashing this out!

ankon: Going to add it as a second commit to this PR then :)

account_test.go
```diff
@@ -26,6 +27,131 @@ import (
 	"time"
 )
 
+// memoryStorage is an in-memory storage implementation with known contents *and* fixed iteration order for List.
+type memoryStorage struct {
+	contents []memoryStorageItem
+}
```
ankon, line 33: FWIW: Apart from needing this for controlling the iteration order of `List`, this might also be faster in general than using an actual file system for unit tests. And, there's good references that say that I/O in unit tests should be banned. :)

account.go
ankon: https://github.com/caddyserver/certmagic/pull/283/commits/97a23b904c7e6dcf820b4aa381eebb1ab8873ef3

account_test.go
mholt: Approved :100:

mholt (approved):
Looks great, thanks so much for the perseverance, @ankon!

PR #286, opened 2024-05-04
Initial implementation of ARI

mholt:
Add ARI (ACME Renewal Information) support.

ARI is currently in draft spec and subject to change: https://datatracker.ietf.org/doc/draft-ietf-acme-ari/03/

For certificates issued with ACME, if the ACME server supports it, CertMagic now considers ARI to trigger renewals.

As usual, the maintenance timer keeps certificates renewed, but now we also refresh ARI if it is time to poll again, and we also use ARI as a determining factor for triggering renewal in addition to the expiration date / lifetime. By default, maintenance routines run every 10 minutes so we will likely not miss an ARI change.

We also store ARI along with the certificate and its other metadata. This means that multiple Caddy instances can share the same ARI data. While we don't strictly synchronize the updating of ARI, it is merely an HTTP request, not a whole transaction with strict CA rate limits, so I don't think we need the complexity there. It is theoretically possible that two instances will update ARI at the same time and both write it to storage, but there's really no harm in that.

We conform to ARI spec recommendations and standards, to the best of my knowledge. Thanks to @aarongable for [answering](https://github.com/aarongable/draft-acme-ari/issues/70) [some](https://github.com/aarongable/draft-acme-ari/issues/71) [questions](https://github.com/aarongable/draft-acme-ari/issues/69) there (though one is still open; not urgent though).

We will ignore ARI if a certificate is dangerously close to the end of its lifetime (last 5% currently) to ensure that buggy ARI implementations on either side do not put the site's availability at risk. Once a certificate is extremely close to expiring, we will renew no matter what ARI says.

I'm still testing this change and finishing it up, but I expect it won't be too much longer before it goes out.

TODO:
- [x] Log when ARI window changes
- [x] Better logging to indicate what triggers a renewal (expiration, ARI window, etc)
- [x] On-demand TLS integration

### DIFF
--- a/certificates.go
+++ b/certificates.go
@@ -18,12 +18,15 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/mholt/acmez/v2/acme"
 	"go.uber.org/zap"
 	"golang.org/x/crypto/ocsp"
 )
@@ -56,6 +59,9 @@ type Certificate struct {
 
 	// The unique string identifying the issuer of this certificate.
 	issuerKey string
+
+	// ACME Renewal Information, if available
+	ari acme.RenewalInfo
 }
 
 // Empty returns true if the certificate struct is not filled out; at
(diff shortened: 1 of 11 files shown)


mholt:
@aarongable @beautifulentropy Would you / anyone from Let's Encrypt be interested in doing a quick once-over to maybe verify the various client aspects of ARI that you're most concerned about as a CA?

config.go
```diff
@@ -869,6 +873,9 @@ func (cfg *Config) renewCert(ctx context.Context, name string, force, interactiv
 				}
 			}
 
+			// the ACME client will use this to tell the server we are replacing a certificate
+			ctx = context.WithValue(ctx, ctxKeyARIReplaces, leaf)
```
oliverpool, line 877: Could this lead to a refusal if the `leaf` was signed by another issuer?
For instance:
- initial certificate signed by ZeroSSL
- upon renewal ZeroSSL is down, so fallback to letsencrypt
- letsencrypt didn't know about this `leaf` certificate, so if I understand https://letsencrypt.org/2024/04/25/guide-to-integrating-ari-into-existing-acme-clients#step-6-indicating-which-certificate-is-replaced-by-this-new-order correcty, it will reject the request

> When Let’s Encrypt processes a new order request featuring a ‘replaces’ field, several important checks are conducted. First, it’s verified that the certificate indicated in this field has not been replaced previously. Next, we ensure that the certificate is linked to the same ACME account that’s making the current request. Additionally, there must be at least one domain name shared between the existing certificate and the one being requested. If these criteria are met and the new order request is submitted within the ARI-suggested renewal window, the request qualifies for exemption from all rate limits. 

This is also suggested by the standard: https://datatracker.ietf.org/doc/draft-ietf-acme-ari/

> Servers SHOULD check that the identified certificate and the current
> New Order request correspond to the same ACME Account and share a
> preponderance of identifiers, and that the identified certificate has
> not already been marked as replaced by a different finalized Order.
> Servers MAY ignore the replaces field in New Order requests which do
> not pass such checks.

(they can't belong to the same ACME account if they are from different issuers)

mholt: Funny puppy being this up, I already have a patch that Flores this locally. I just need to test it and commit it.

mholt:
Latest commit ensures CA is the same before setting `Replaces` on the ACME order.

We also don't set `Replaces` on the 2nd retry (3rd attempt) of ACME transactions. This is in case the reason for the failed order is a rejected `Replaces` value. I chose 2nd retry in case the first failure is maybe sporadic, the first retry will fix it. 2nd retry is early on in the process but then if that fails too, then we know it's not related to ARI, so we can just keep retrying normally after that and still gain the benefits of updating the server state when it eventually does succeed.

I've also complexified the logic for deciding whether a certificate needs renewal. Before this PR, we just checked the expiration date, the current time, and the configured lifetime ratio. Now this change considers ARI if it is available, but the logic is nuanced to ensure maximum reliability and flexiblity:

- If the ARI already has a selected time, use that. **We do not also require a window to exist in the metadata.** This allows sysadmins to manually schedule renewals independent of ARI which sounds useful. All they have to do is set the value in the JSON accordingly.
- If there is an ARI window, but no selected time (for some reason?), then select a time. It's ephemeral (we don't store this generated random time) but that's OK. We just need a random time in the window.
- If the selected time says it's not time to renew, we still check the expiration date to make sure a bug or state inconsistency somewhere doesn't just cause us to sit back and let the cert expire. Always check the expiration date. It has to be in the last 1/20 of the cert lifetime to supercede ARI here.
- Otherwise, if there's no ARI or selected renewal time, just check the expiration against the current time and the configured renewal window ratio. This is the same as the old behavior.
- If it still isn't deemed time to renew, do one last check for an emergency renewal, in the last, I dunno, 1/50th the cert's lifetime, OR the last few runs of the maintenance routine before the cert expires. This should help ensure that no matter what (mis)configuration there is, the certificate will be renewed as long as the server is running.

Still need to integrate with on-demand TLS and log when we detect a change to the ARI window.

config.go
mholt: Thanks for the analysis btw, great writeup/catch!

mholt:
The latest commit now checks ARI with on-demand TLS and when loading certs from storage (rather than just checking expiration date).

I've also added logging when a window change is detected, including the explanation URL.

I've also ensured that any updated ARI will trigger a renewal of the certificate if the new ARI window means that the certificate is to be renewed immediately.

After testing several scenarios with this code, I'm confident enough to merge it in, but I want more testing in the field before it gets released in a Caddy 2.8 production release.

certificates.go
```diff
@@ -67,10 +73,106 @@ func (cert Certificate) Empty() bool {
...
+	// routine to check for renewals, to accommodate both exceptionally long and short
+	// cert lifetimes
+	if currentlyInRenewalWindow(leaf.NotBefore, expiration, 1.0/50.0) ||
+		time.Until(expiration) < cfg.certCache.options.RenewCheckInterval*5 {
+		logger.Warn("certificate is in emergency renewal window; expiration imminent",
+			zap.Duration("remaining", time.Until(expiration)))
+		return true
+	}
+
+	return false
```
oliverpool, line 175: The comments look sensible. Maybe some unit test of thie `certNeedsRenewal` function would be beneficial as well? (to ensure that any change does not break expected behavior)

mholt:
I'm going to merge this so I can get it some practice in the field, but feedback/review is still welcomed if anyone desires to; we can still make fixes in follow-up commits/PRs.

PR #287, opened 2024-05-07
Apply DefaultServerName more broadly during handshake

mholt:
Currently, we only use `DefaultServerName` for choosing a certificate. This doesn't make a lot of sense, because if that is set to a name that doesn't have a cert managed for it, it will not be of much use because we'd use the IP address to fill in an empty ServerName when doing logic, but only using DefaultServerName when loading the certs. So things didn't line up sometimes.

This change makes it so that DefaultServerName will always be used as the ServerName when it was empty, meaning that the IP address will never be used as the "name". This is more expected behavior. (So obviously, if you WANT to serve your IP address over TLS, then don't set DefaultServerName.)

This is being deployed soon to a production environment but is so far working well in staging. Likely to be merged into mainstream CM.

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -120,7 +120,7 @@ func (cfg *Config) getCertificateFromCache(hello *tls.ClientHelloInfo) (cert Cer
 			}
 		}
 
-		// fall back to a "default" certificate, if specified
+		// use a "default" certificate by name, if specified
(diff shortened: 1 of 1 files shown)


mholt:
Awesome!! I'll merge it in.

PR #289, opened 2024-05-15
downgrade minimum Go version

mohammed90:
This currently does not work because it requires caddyserver/zerossl#1 to be merged first. That dependency forced this module to upgrade its minimum Go version. That is not needed, and it's causing pains due to mis-synchronization with other parts of Caddy ecosystem. Once caddyserver/zerossl#1 is merged, the dependency here can be updated to use the next patch release of the zerossl package, then do `go mod tidy` and set this PR to be ready.

Depends on caddyserver/zerossl#1

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/caddyserver/certmagic
 
-go 1.22.0
+go 1.21.0
 
 toolchain go1.22.2
 
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.21.0
 toolchain go1.22.2
 
 require (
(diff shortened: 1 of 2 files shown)


mholt:
I've pushed v0.1.3 of the zerossl package which has the lower go version in go.mod; do we need to upgrade that dep in this PR?

mohammed90:
>  do we need to upgrade that dep in this PR?

Yes. I'm mobile right now. If you can do it, go for it. Otherwise, it's just as upgrade of zersl then `go mod tidy`.

mholt:
I get "permission denied" when trying to push

mholt (approved):
Thanks for fixing the permission issue!

Issue #268, opened 2024-02-08
Ability to disable logs with `no information found to solve challenge for identifier`


mbardelmeijer:
## What would you like to have changed?
We use certmagic's OnDemand method to allow and disallow specific domains. If a domain is disallowed, we sometimes still see logs about this domain like:

```
{"level":"error","ts":1707408179.1513515,"caller":"certmagic@v0.20.0/httphandler.go:76","msg":"looking up info for HTTP challenge","component":"acme_issuer","host":"XXXXXX","remote_addr":"XXXXXX","user_agent":"cert-manager/v1.5.3 (clean)","error":"no information found to solve challenge for identifier: XXXXXX","stacktrace":"github.com/caddyserver/certmagic.(*ACMEIssuer).distributedHTTPChallengeSolver\n\t/go/pkg/mod/github.com/caddyserver/certmagic@v0.20.0/httphandler.go:76\ngithub.com/caddyserver/certmagic.(*ACMEIssuer).HandleHTTPChallenge\n\t/go/pkg/mod/github.com/caddyserver/certmagic@v0.20.0/httphandler.go:62\nmain.(*httpServer).serveRequest\n\t/github/workspace/http.go:226\nmain.newHttpServerRateLimiter.func1.1\n\t/github/workspace/http.go:157\nnet/http.HandlerFunc.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2136\nnet/http.HandlerFunc.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2136\nnet/http.(*ServeMux).ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2514\ngithub.com/bugsnag/bugsnag-go/v2.Handler.func1\n\t/go/pkg/mod/github.com/bugsnag/bugsnag-go/v2@v2.2.0/bugsnag.go:182\nnet/http.HandlerFunc.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2136\nnet/http.serverHandler.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2938\nnet/http.(*conn).serve\n\t/usr/local/go/src/net/http/server.go:2009"}
```

## Why is this feature a useful, necessary, and/or important addition to this project?
This allows us to better filter out these error messages and keep our general logs clean.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
[...]

PR #290, opened 2024-05-24
Move `looking up info for HTTP validation` to warn level

mbardelmeijer:
This PR moves `looking up info for HTTP {challenge,validation}` from an error to a warning level.

The reasoning for this is as follows: we see with our certmagic deploy that the `.well-known/acme-challenge/xxx` paths are automatically checked by old issuers that still think they are managing the domain. These acme-challenge paths are then checked by certmagic to solve the challenge, but as it's an unknown challenge token, this fails.

It's still a relevant message to log, although not at the error level, IMHO.

Fixes https://github.com/caddyserver/certmagic/issues/268


francislavoie:
Makes sense to me. I've seen these in the wild on the Caddy forums on occasion, and I agree they're not errors, they should be warnings.

mholt (approved):
Okay, I think that's reasonable. Thanks!

Issue #296, opened 2024-06-21
Do calls to storage Load need to be Lock protected?


elee1766:
by contract, do calls to load need to be protected by the storage lock?

from what I see, the lock is used only to protect for simultaneous writes. it's not used to lock/protect the reader if a writer is currently writing. 

is the caller supposed to take the Lock before reading the certificate? if not, are storage Load implementations supposed to protect for concurrent read/write access?

elee1766:
Filestore uses os.WriteFile

```
// WriteFile writes data to the named file, creating it if necessary. 

// If the file does not exist, WriteFile creates it with permissions perm (before umask); 

// otherwise WriteFile truncates it before writing, without changing permissions. 

// Since WriteFile requires multiple system calls to complete, a failure mid-operation 

// can leave the file in a partially written state. 

func WriteFile(name string, data []byte, perm FileMode) error { 

	f, err := OpenFile(name, O_WRONLY|O_CREATE|O_TRUNC, perm) 

	if err != nil { 

		return err 

	} 

	_, err = f.Write(data) 

	if err1 := f.Close(); err1 != nil && err == nil { 

		err = err1 

	} 

	return err 

}
```

if reader reads while writefile is running. it could read after open file, which truncates the file, so it will read empty.

so I think there is a race and protection is needed

elee1766:
https://github.com/containerd/containerd/blob/main/pkg%2Fatomicfile%2Ffile.go
maybe can do what containerd does

mholt:
> by contract, do calls to load need to be protected by the storage lock?

No; this synchronization should be handled primitively, natively/internally, within the storage mechanism. The user shouldn't have to lock for every other storage operation.

> from what I see, the lock is used only to protect for simultaneous writes. it's not used to lock/protect the reader if a writer is currently writing.

Locking is for user-scope purposes, like renewing certificates or something. So it can protect reads and writes, but what it's really for is to sync a certain operation or task that may involve multiple reads or writes.

> if reader reads while writefile is running. it could read after open file, which truncates the file, so it will read empty.

It sounds like you know a lot more about the primitives of file systems than I do :sweat_smile: Especially after our conversation in slack. We can implement improvements to the FileStorage if needed! Is it a matter of calling Sync() in more places?

elee1766:
> > by contract, do calls to load need to be protected by the storage lock?
> 
[...]

PR #300, opened 2024-07-06
fix: use a tempfile to write files in filestorage. 

elee1766:
discovered in #296 

there is a currently a race between file deletion and rewrite which can cause an empty read. 

copying the method used by containerd, https://github.com/containerd/containerd/blob/main/pkg%2Fatomicfile%2Ffile.go the problem can be solved via a flushing a temporary file

### DIFF
--- a/filestorage.go
+++ b/filestorage.go
@@ -27,6 +27,8 @@ import (
 	"path/filepath"
 	"runtime"
 	"time"
+
+	"github.com/caddyserver/certmagic/internal/atomicfile"
 )
 
 // FileStorage facilitates forming file paths derived from a root
@@ -82,12 +84,30 @@ func (s *FileStorage) Store(_ context.Context, key string, value []byte) error {
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(filename, value, 0600)
+	fp, err := atomicfile.New(filename, 0o600)
+	if err != nil {
+		return err
+	}
+	_, err = fp.Write(value)
+	if err != nil {
+		// cancel the write
+		fp.Cancel()
+		return err
+	}
+	// close, thereby flushing the write
+	return fp.Close()
 }
 
 // Load retrieves the value at key.
 func (s *FileStorage) Load(_ context.Context, key string) ([]byte, error) {
-	return os.ReadFile(s.Filename(key))
+	// i believe it's possible for the read call to error but still return bytes, in event of something like a shortread?
+	// therefore, i think it's appropriate to not return any bytes to avoid downstream users of the package erroniously believing that
+	// bytes read + error is a valid response (it should not be)
+	xs, err := os.ReadFile(s.Filename(key))
+	if err != nil {
+		return nil, err
+	}
+	return xs, nil
 }
 
 // Delete deletes the value at key.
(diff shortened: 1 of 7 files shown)


mholt:
Thanks for the PR!

I don't understand much about this (you have been very helpful explaining it in Slack so far, which I appreciate) -- is this change ready to go? Does it work for all the platforms it is supposed to work on (and it doesn't break on others)?

I will give it a closer look ASAP :)

elee1766:
yes this is ready. it should work fine and shouldn't break anything on windows.

filestorage.go
```diff
@@ -82,12 +84,30 @@ func (s *FileStorage) Store(_ context.Context, key string, value []byte) error {
...
 	}
-	return os.WriteFile(filename, value, 0600)
+	fp, err := atomicfile.New(filename, 0o600)
+	if err != nil {
+		return err
+	}
+	_, err = fp.Write(value)
+	if err != nil {
+		// cancel the write
+		fp.Cancel()
```
mholt, line 94: Why doesn't the fp automatically cancel the write when there's an error?

elee1766: cancel is a special call which deletes the temp file.

the fp to the tempfile can't know to delete itself when it closes.

mholt:
One more quick question -- how do you think this will work with NFS mounts? We have some people using it with NFS, though I don't recommend it because I know NFS already has some issues related to O_EXCL and possibly other things; do you think this will make the problem worse or better? (Or neither)

elee1766:
> One more quick question -- how do you think this will work with NFS mounts? We have some people using it with NFS, though I don't recommend it because I know NFS already has some issues related to O_EXCL and possibly other things; do you think this will make the problem worse or better? (Or neither)

it should be better or the same. 

NFS mounts have more lag, they are maybe more likely to run into the race condition that this fixes. 

but also, NFS has so many issues for stuff like this that I can't say for sure it will anything better. I don't know if this will fix issues with o_excl. I don't use it and it's fundamentally broken on NFS so I doubt it will.

internal/testutil/readme.md
```diff
@@ -0,0 +1,3 @@
+# testutil
+
+some testing functions copied out of github.com/stretchr/testify, which were originally used by atomicfile
```
mholt, line 3: This is fine -- but it's MIT licensed code so we should probably copy and paste the license (which contains the copyright) into this readme file, just in case.

mholt (approved):
Let's give this a shot. Thanks for being willing to help maintain this!

internal/testutil/readme.md
elee1766: done

mholt (approved):

ghasminia:
Is this in version 2.8.4?
Updated 9 days ago:
https://hub.docker.com/layers/library/caddy/2.8.4/images/sha256-33240ceb04db42b1e8fb24be92fab799c83234302fbfcfc0d7b157e477fb832a?context=explore

If not, when can I use it? I am managing about 3000 domains with caddy and I have this problem seriously

mholt:
Very soon. When I get back from Europe I'll tag a new version. You can build yourself though until then, I'll try to post a sample command as soon as I can!

PR #311, opened 2024-09-15
refactor: replace the file storage logger with the default logger

jinrenjie:
I use certmagic in my project, it's great, but I had trouble collecting logs, so I want to replace `log` with `Default.Logger` to make the log style uniform!


mholt:
Thanks; I _suppose_ this is better than using the std lib logger, as it's the "wrong" logger either way (these places don't have a config available to them for us to get the "correct" logger).

Although, `Default` is intended as a template and shouldn't be used directly. Plus, it could be modified to set `Default.Logger` to `nil`. (:boom:) We should probably refer to `defaultLogger` directly if possible. Would that work?

mholt:
>  In addition, I saw in the code that if Default.Logger is nil, defaultLogger can be assigned to Default.Logger

That's only the case if a new instance of a Config is made by using Default as an input (like a template).

`defaultLogger` is the only zap logger guaranteed to be properly set (non-nil).

filestorage.go
```diff
@@ -311,15 +309,15 @@ func keepLockfileFresh(filename string) {
 		if err := recover(); err != nil {
 			buf := make([]byte, stackTraceBufferSize)
 			buf = buf[:runtime.Stack(buf, false)]
-			log.Printf("panic: active locking: %v\n%s", err, buf)
+			Default.Logger.Sugar().Panicf("active locking: %v\n%s", err, buf)
```
francislavoie, line 312: I think we intentionally don't want to panic here but maybe I'm wrong. A panic would take down the server, we intentionally `recover()` to avoid that from happening, then just log that it happened and keep going.

jinrenjie: Yes, Panic should not be used here, I have replaced Panicf with Errorf.

jinrenjie:
@mholt I have replaced Default.Logger with defaultLogger.

mholt:
Thanks -- sorry I've been backlogged. This is still on my list just FYI :)

PR #316, opened 2024-10-24
ZeroSSLIssuer: Make PollInterval configurable

aplr:
This PR adds the `PollInterval` option to the `ZeroSSLIssuer`, which lets users adjust the polling interval for certificate retrieval when using the ZeroSSL issuer.

If no `PollInterval` is configured (or a value of `0` is set), the default poll interval of 5 seconds is used.


mholt (approved):
Thanks! Nice enhancement.

PR #317, opened 2024-10-30
Remove obsolete constants

ankon:
The validation path prefix is likely now `zerosslHTTPValidationBasePath`, and the API base here is broken (`"https://" + somethingThatIncludesHTTPSAsWell`).


mholt (approved):
Ah thanks! :man_facepalming:

PR #318, opened 2024-11-02
Add a fast path for the DefaultCertificateSelector

ankon:
In cases where we only have a single certificate to choose from we will anyways pick that certificate, regardless of whether the certificate is supported by the client or is expired.

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -231,9 +231,16 @@ func (cfg *Config) selectCert(hello *tls.ClientHelloInfo, name string) (Certific
 // otherwise it returns an expired certificate that the client supports,
 // otherwise it just returns the first certificate in the list of choices.
 func DefaultCertificateSelector(hello *tls.ClientHelloInfo, choices []Certificate) (Certificate, error) {
+	if len(choices) == 1 {
+		// Fast path: There's only one choice, so we would always return that one
+		// regardless of whether it is expired or not compatible.
+		return choices[0], nil
+	}
 	if len(choices) == 0 {
 		return Certificate{}, fmt.Errorf("no certificates available")
 	}
+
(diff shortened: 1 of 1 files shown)


handshake.go
```diff
@@ -231,9 +231,16 @@ func (cfg *Config) selectCert(hello *tls.ClientHelloInfo, name string) (Certific
...
+		// Fast path: There's only one choice, so we would always return that one
+		// regardless of whether it is expired or not compatible.
+		return choices[0], nil
+	}
 	if len(choices) == 0 {
 		return Certificate{}, fmt.Errorf("no certificates available")
 	}
+
+	// Slow path: There are choices, so we need to check each of them.
 	now := time.Now()
```
ankon, line 244: I didn't check, but: There might be a good opportunity to split the slow path out into a separate function, so that the fast path can be inlined.

mholt (approved):
Thanks for this optimization, makes sense!

At first I read the comment, and was like, "No, wait, we shouldn't serve expired or incompatible certs." But then I realized that we already do this below (`best := choices[0]`) and remembered why... it's generally best practice to serve _any matching certificate you have_ in response to a ClientHello because it can help with troubleshooting. Like if you serve an expired cert, you can instantly know that something about its renewal failed (for a long time!)... or if it's an incompatible cert, you can verify that everything on the server is correct and that the client program just needs to get over itself and support it :sweat_smile:

Anyway, thanks again for this.

handshake.go
mholt: Should be doable. Lemme catch up on this next Caddy beta release and then I/we can explore that as a follow-up PR. I'd be curious if the inlining actually happens and if so does the performance improve?

PR #320, opened 2024-11-08
Quiet OCSP warnings if the cert has a short lifetime

francislavoie:
In Caddy, we often get warnings like these when using internally issued certs:

```
{"level":"warn","ts":1731027618.3063543,"logger":"tls","msg":"stapling OCSP","error":"no OCSP stapling for [localhost]: no OCSP server specified in certificate","identifiers":["localhost"]}
```

These generally have a very short lifetime and for that reason warning about them not having OCSP configured is just noise, because OCSP on short certs isn't too useful.

I added a `Lifetime()` method to help here, I used `expiresAt` for the "1s resolution of ASN.1 UTCTime/GeneralizedTime", hope that's correct.

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -98,7 +98,7 @@ func stapleOCSP(ctx context.Context, ocspConfig OCSPConfig, storage Storage, cer
 		if ocspErr != nil {
 			// For short-lived certificates, this is fine and we can ignore
 			// logging because OCSP doesn't make much sense for them anyway.
-			if cert.Lifetime() > 7*24*time.Hour {
+			if cert.Lifetime() < 7*24*time.Hour {
 				return nil
 			}
 			// There's nothing else we can do to get OCSP for this certificate,
--- a/ocsp.go
+++ b/ocsp.go
@@ -93,11 +93,16 @@ func stapleOCSP(ctx context.Context, ocspConfig OCSPConfig, storage Storage, cer
 	// then we need to request it from the OCSP responder
 	if ocspResp == nil || len(ocspBytes) == 0 {
 		ocspBytes, ocspResp, ocspErr = getOCSPForCert(ocspConfig, pemBundle)
+		// An error here is not a problem because a certificate
+		// may simply not contain a link to an OCSP server.
 		if ocspErr != nil {
-			// An error here is not a problem because a certificate may simply
-			// not contain a link to an OCSP server. But we should log it anyway.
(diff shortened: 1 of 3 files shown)


mholt (commented):
Looks good, overall!

One thought I had: might it be a good idea to only stifle the warnings for the specific message of missing OCSP information (in short-lived certs)?

Because, if for some reason a short-lived cert had OCSP, you'd probably still want to know it. But I think specifically what we want to do is make it completely silent/OK if a short-lived cert specifically doesn't support OCSP. But if there was some sort of other error, like a network issue, we'd want to know it.

What do you think? The logic of calling `Lifetime()` would probably move into `stapleOCSP` and if the lifetime was short, we'd just set the err to nil and return (a no-op).

francislavoie:
How's this?

ocsp.go
```diff
@@ -93,11 +93,16 @@ func stapleOCSP(ctx context.Context, ocspConfig OCSPConfig, storage Storage, cer
...
 	if ocspResp == nil || len(ocspBytes) == 0 {
 		ocspBytes, ocspResp, ocspErr = getOCSPForCert(ocspConfig, pemBundle)
+		// An error here is not a problem because a certificate
+		// may simply not contain a link to an OCSP server.
 		if ocspErr != nil {
-			// An error here is not a problem because a certificate may simply
-			// not contain a link to an OCSP server. But we should log it anyway.
+			// For short-lived certificates, this is fine and we can ignore
+			// logging because OCSP doesn't make much sense for them anyway.
+			if cert.Lifetime() > 7*24*time.Hour {
```
mholt, line 101:
```suggestion
			if cert.Lifetime() < 7*24*time.Hour {
```

mholt (changes requested):
Yes! That's it! Almost, I think :smile:

ocsp.go
francislavoie: Oh yeaah right, just copied the condition but forgot to flip it. Booleans are hard.

mholt (approved):
:star: :star2: :star:

mholt:
Oh, lol, whodathunkit, we have a test for this: `ocsp_test.go:151: expected error "no OCSP stapling for [ocsp test certificate]: no URL to issuing certificate" but got %!q(<nil>)`

francislavoie:
Oh I guess this broke a test lmao

mholt (approved):

mholt:
Thanks a bunch!

PR #337, opened 2025-03-28
File lock changes

elee1766:
this changes the writing of the lock file to be a write and rename instead of an open and write

this introduces a possible race, as the code originally relied on the exclusive mode to write the file.

so fix one race to introduce another. lol


mholt:
So, I am not sure if introducing one race to fix another is a great solution tbh. Sorry for being egregiously slow to respond to this... do you think we should close this?

PR #345, opened 2025-07-01
acmeissuer: LE now supports IP certs

mohammed90:
See: https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate/


mholt (commented):
Thanks!

This is probably the way to go, although it's worth noting that LE only issues IP certs using the `shortlived` profile.

I don't want to get into the business of duplicating CA-specific requirements, though -- it's mainly a high-level sanity check here, so, I think we can merge this and give it a try. It'll be on the user to use the shortlived profile, similar to how it's on the user to enable DNS challenge for wildcard certificates.

PR #347, opened 2025-08-01
Implement precise lock lease renewal for storage backends that support lease renewal.

zoltan-frm:
- Added:`LockLeaseRenewer` interface to support renewing lock leases. `renewLockLease` function to extend lease duration based on retry attempts.
- Updated certificate management functions to renew lock leases during long-running retry loops.
- Added tests for lease renewal functionality in `config_test.go` to ensure correct behavior with and without lease support.


zoltan-frm:
Hi @mholt , this has been in production for a week with us. Although the number of ACME requests have not been dramatically reduced, we would be happy to have it, as it is still beneficial to avoid races across instances.

mholt (approved):
Sounds good, thanks! Will merge this in.

Am working on the "stateless" solving next.

PR #351, opened 2025-08-27
Add `.test` to be classified as an "internal-only hostname"

jaydrogers:
## What this PR does
- This adds `.test` to be classified as an "internal" TLD

## Background
The [Caddy docs](https://caddyserver.com/docs/automatic-https#hostname-requirements) specifically reference the following TLDs as "internal":

- .localhost
- .local
- .internal
- .home.arpa

## Why this change is proposed
1. `.test` is a domain that is reserved specifically for testing DNS and would never be resolved by a root TLD server (causing ACME challenges to never resolve anyways)
2. Many developers use `.test` for local development

## How this improves the product if merged
It can improve developer experience allowing `myapp.test` to be set as a hostname without having Caddy trying to run an ACME challenge for trusted SSL.

**Example Caddyfile**
```caddyfile
myapp.test {
  # My Caddy configuration
}
```

The expected result after this change would trigger the "internal" SSL generation instead of trying to obtain one through Let's Encrypt, etc.

## ⚠️ Warning: I'm not a Go developer
I was able to form this PR by following the structure of this previous commit: https://github.com/caddyserver/certmagic/commit/66685874a86f20f433d38d00e2ab8a9250522b0d, which was the solution to this issue https://github.com/caddyserver/caddy/issues/5800

Please feel free to scrutinize or improve this PR as needed. I just put this together to make it easy as possible for the maintainers as long as my syntax is true 🤠

## Related docs change
If this change is accepted, I also prepared a PR on the docs site to document this change:

- https://github.com/caddyserver/website/pull/491

## References
Please see official documentation below to explain more on `.test`.
- [IETF RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606#section-2) - TLDs for testing
- [IETF RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761#section-6.2) - Domain Name Reservation Considerations for "test."

Thanks for your support and maintenance of this project 🙏


jaydrogers:
Ha! Thanks for sharing the XY problem. 😃

To your comment:

> I think this approach would be more of a band-aid and will only serve a few use cases, whereas there are typically many things that need adjustment from dev to production.

My use case really isn’t in scope of this discussion though, I was just sharing on what motivated me.

This PR is all about 👉 **What hostnames qualify for publicly-trusted certificates?**

My proposal is to add `.test` because from my understanding of Caddy (obviously I’m the novice here), I could not see a use case where someone needed an automated and trusted SSL certificate on an external server with `.test`. To my understanding, it would not even work because ACME validations would fail because `.test` would be an _internal_ configuration.

Since I saw a more common use cast of `.test` being in local development, I created the PR to have it act like `.localhost`.

Food for thought while you think this through.

I appreciate the conversation. Have a good weekend ✌️

timelordx:
> ## Where I ran into a challenge
> The problem was that if I want to take the image from Development to Production, all I should need to do is change the environment variable:
> 
> Environment	`${SERVER_NAME}` Value
> Development	example.dev.test
> Production	example.com
> ## My expected result
> I was hoping the user could set `${SERVER_NAME}` to whatever the domain was and it would:
> 
> 1. Self-sign a certificate in development (because of `.test`)
> 2. Automatically generate a trusted certificate in production (because of `.com`)
> 
> ### Why this wont work
> In order to use `.test` and `.com` across the same app, I would have to create two blocks:
> 
> #### Development (`SERVER_NAME=example.dev.test`)
> ```diff
> {$SERVER_NAME} {
> +       internal
> 	# My config
> }
> ```
> 
> #### Production (`SERVER_NAME=example.com`)
> Then I would need a separate block for production and remove `internal`, because I want automatic via ZeroSSL or Let's Encrypt.
> 
> ```diff
> {$SERVER_NAME} {
> -       internal
> 	# My config
> }
> ```


This can easily be addressed via another environment variable, for example, `${TLS_DIRECTIVE}`, with the common stanza for both environments:
```
{$SERVER_NAME} {
	{$TLS_DIRECTIVE}
	# My config
}
```

#### Development
```
SERVER_NAME=example.dev.test
TLS_DIRECTIVE='tls internal'
```
#### Production
```
SERVER_NAME=example.com
TLS_DIRECTIVE=''
```

PR #363, opened 2026-01-01
chore: use math/rand/v2 instead of math/rand

WeidiDeng:
Similar to [7413](https://github.com/caddyserver/caddy/pull/7413).

### DIFF
--- a/config.go
+++ b/config.go
@@ -1249,7 +1249,7 @@ func (cfg *Config) checkStorage(ctx context.Context) error {
 		contents[i+6] = byte(v >> 48)
 		contents[i+7] = byte(v >> 56)
 	}
-	err = cfg.Storage.Store(ctx, key, contents)
+	err := cfg.Storage.Store(ctx, key, contents)
 	if err != nil {
 		return err
 	}


config.go
```diff
@@ -1236,7 +1236,7 @@ func (cfg *Config) checkStorage(ctx context.Context) error {
 	}
 	key := fmt.Sprintf("rw_test_%d", weakrand.Int())
 	contents := make([]byte, 1024*10) // size sufficient for one or two ACME resources
-	_, err := weakrand.Read(contents)
+	_, err := rand.Read(contents)
```
mholt:
```suggestion
	_, err := weakrand.Read(contents)
```

Should be `weakrand`, right?

certificates.go
```diff
@@ -21,7 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math/rand"
+	"math/rand/v2"
```
mholt: I wonder if we should import this as `weakrand` for consistency with other files. It's a convention that helps ensure readers know the package is not strong enough for cryptographic use. But then again I guess linters do this too, so I dunno.

mholt (changes requested):
Thanks Weidi! I just noticed one thing I wanted to check on.

Issue #361, opened 2025-11-27
HTTP 409 "Conflict" when the certificate identified by 'replaces' has already been replaced


vmihailenco:
## What version of the package are you using?

v0.25.0

## What are you trying to do?

I've been hit by https://github.com/ietf-wg-acme/acme-ari/issues/56 . I see the comment on the thread indicating that nothing can be done on the client, but the error message indicates otherwise and there PRs like [this](https://github.com/go-acme/lego/pull/2475/files).

## What steps did you take?

I had to manually remove existing certificates from the storage to let the renewal proceed.

## What did you expect to happen, and what actually happened instead?

The certificate must be renewed. Instead, the certificate expired and could not renewed.

## How do you think this should be fixed?

Either https://github.com/go-acme/lego/pull/2475/files or just deleting broken certificates from the storage.

## Please link to any related issues, pull requests, and/or discussion

https://github.com/ietf-wg-acme/acme-ari/issues/56
https://github.com/go-acme/lego/pull

## Bonus: What do you use CertMagic for, and do you find it useful?

👍

vatsaljrathod:
We got the same issue. I am sharing the logs(i updated the actual certificate host with abc.xyz), might be useful. 

info	maintenance	started background certificate maintenance	{"cache": "0xc0001c0180"}
info	certificate needs renewal based on ARI window	{"subjects": ["*.abc.xyz"], "expiration": 1768505981, "ari_cert_id": "jw0TovYuftFQbDMYOF1ZjiNykco.Bb2fOSS1ZG9wIviLW3hG60vQ", "next_ari_update": 1765972364.6836352, "renew_check_interval": 600, "window_start": 1765837425, "window_end": 1765992874, "selected_time": 1765844188, "renewal_cutoff": 1765843588}
info	renew	acquiring lock	{"identifier": "*.abc.xyz"}
info	renew	lock acquired	{"identifier": "*.abc.xyz"}
info	renew	renewing certificate	{"identifier": "*.abc.xyz", "remaining": 2537398.605025773}
info	waiting on internal rate limiter	{"identifiers": ["*.abc.xyz"], "ca": "https://acme-v02.api.letsencrypt.org/directory", "account": "support@abc.xyz"}
info	done waiting on internal rate limiter	{"identifiers": ["*.abc.xyz"], "ca": "https://acme-v02.api.letsencrypt.org/directory", "account": "support@abc.xyz"}
[...]

PR #364, opened 2026-01-07
Only indicate the replaced cert if the ARI suggested window is not in the past

oliverpool:
This should help address #361:
instead of always indicating the replaced cert, only fill this field if the ARI suggested window is not over.

This is not a perfect solution (ideally the `alreadyReplaced` response should be explicitly handled), however it should prevent having the current case of unrenewable certificates.

### DIFF
--- a/config.go
+++ b/config.go
@@ -858,7 +858,7 @@ func (cfg *Config) renewCert(ctx context.Context, name string, force, interactiv
 		}
 
 		// check if renew is still needed - might have been renewed while waiting for lock
-		timeLeft, leaf, needsRenew := cfg.managedCertNeedsRenewal(certRes, false)
+		timeLeft, replacesCert, needsRenew := cfg.managedCertNeedsRenewal(certRes, false)
 		if !needsRenew {
 			if force {
 				log.Info("certificate does not need to be renewed, but renewal is being forced",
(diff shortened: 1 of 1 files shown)


oliverpool:
> Hmm, I am not sure if the time relative to the ARI window's end is the right condition for whether to set replaces.

Currently the `replace` field is only set if ARI is enabled. My PR tightens this condition to “if ARI is set, supported by the server and not in the past” (the server-support is needed according to https://github.com/ietf-wg-acme/acme-ari/blob/main/rfc9773.md#extensions-to-the-order-object “Clients SHOULD NOT include this field if the ACME server has not indicated that it supports this protocol”)

---

My understanding of (one) ARI usecase is to be able to quickly rotate certificates after an issuer had some security issues. In this case, it might be interesting for the issuer to know when the “wrong” certificate has been replaced, to actively revoke it (hence the "replace" field).

In this (rare) case, the certmagic client will have a big problem (revoked certificate, but the new one didn’t make it somehow). 

In all cases, if the ARI window is in the past, it likely means that the attempts made during the ARI window failed. So the issuer might not get the replacement info, but at least the client will not be stuck with an expiring certificate.

oliverpool:
Another approach would be to catch this specific error code and retry immediately without the replace field (in this case).

PR #376, opened 2026-04-11
fix: Normalization IPv6 addresses for ACME challenge

u5surf:
* subsumed https://github.com/caddyserver/caddy/issues/7399

When the caddy server uses ACME with IPv6 addresses, it had been with bracket as `[::1]` in standard Go HTTP Request host. The host name evaluation is used without bracket IPv6, it must be eliminated bracket. And their operations should be handled within this module as this comment.

https://github.com/caddyserver/caddy/pull/7619#issuecomment-4227356458


mholt:
Thanks; sorry I forgot about this PR earlier. Looks good, let's try it!

mholt (approved):

PR #377, opened 2026-04-21
Fix HTTP-01 challenge for IPv6 literal addresses

nekohasekai:
## What changed

Strip square brackets from bare IPv6 literals in `hostOnly()` when `net.SplitHostPort` returns an error.

## Why

For HTTP-01 validation against an IPv6 literal, the CA sends a `Host` header like `[2001:db8::1]` without a port.

`hostOnly()` returns that bracketed value, while `challenge.Identifier.Value` holds the bare IPv6 address, so the host check fails.

## Validation

- `go test ./...`

### DIFF
--- a/certificates.go
+++ b/certificates.go
@@ -651,6 +651,9 @@ func isInternalIP(addr string) bool {
 func hostOnly(hostport string) string {
 	host, _, err := net.SplitHostPort(hostport)
 	if err != nil {
+		if len(hostport) > 2 && hostport[0] == '[' && hostport[len(hostport)-1] == ']' {
+			return hostport[1 : len(hostport)-1]
+		}
 		return hostport // OK; probably had no port to begin with
 	}
 	return host


mholt:
Huh... is that even valid form? I thought `[]` only should be used when there's a port. And there's no ports in ACME challenge identifier values. Where are you seeing this?

nekohasekai:
Yes, it's valid form — see RFC 9112 §3.2: `Host = uri-host [ ":" port ]`, and `uri-host` is defined in RFC 3986 §3.2.2 where `IP-literal = "[" IPv6address "]"`. The brackets are part of the IP-literal grammar, not the port separator, so the port is independently optional.

It's exactly what Boulder produces. [va/http.go](https://github.com/letsencrypt/boulder/blob/main/va/http.go#L445-L464) wraps IPv6 identifiers in brackets and then constructs `url.URL{Scheme: "http", Host: host, Path: path}` without a port (httpPort is only handed to the dialer). When Go's `net/http` serializes that to the wire, the Host header is literally `[2001:db8::1]` — no port.

Reproducer:

```go
req, _ := http.NewRequest("GET", "http://[2001:db8::1]/.well-known/acme-challenge/x", nil)
// req.Host == "[2001:db8::1]"
// wire: "Host: [2001:db8::1]"
```

Downstream this hits `hostOnly()` in certificates.go:651. `net.SplitHostPort("[2001:db8::1]")` returns `missing port in address`, so it falls through to return the bracketed input, and `strings.EqualFold("[2001:db8::1]", "2001:db8::1")` fails the DNS-rebinding check in `solveHTTPChallenge`.

mholt (approved):
Ok. Fair enough I guess, thank you

PR #382, opened 2026-05-09
Preserve DNS provider metadata for DNS-01 cleanup

### LINKED ISSUE #381: Preserve DNS provider metadata for DNS-01 cleanup
CertMagic currently stores `results[0].RR()` after `AppendRecords` succeeds during DNS-01 presentation. This preserves the portable DNS record data, but discards provider-specific metadata from the returned `libdns.Record`, such as `ProviderData`.

That metadata can be important for cleanup. Some libdns providers receive an internal record ID from the provider API when creating a record and store it in `ProviderData`. If CertMagic discards it, providers that delete by internal ID may need to rediscover the record later, usually by listing zone records and matching on name/type/value, before they can delete by provider ID.

This is less robust than preserving the record returned by `AppendRecords`:

- It can require broader token permissions, such as `list-records`, even when create/delete-by-ID would otherwise be sufficient.
- It adds an avoidable provider API call during cleanup.

CertMagic should preserve the full returned `libdns.Record` while the challenge is active, and use `record.RR()` only where portable DNS fields are needed for propagation checks, logging, and memory lookup. Cleanup can then pass the original returned record back to `DeleteRecords`.

This would not make CertMagic depend on provider-specific IDs. Providers without metadata still receive the same standard DNS data through `Record.RR()`, and their existing fallback behavior continues to work. It simply avoids discarding provider metadata when it is available.

bjornmp:
## Summary

This change stores the full `libdns.Record` returned by `AppendRecords` while a DNS-01 challenge is active, instead of reducing it to `record.RR()` immediately.

The portable DNS fields are still accessed through `record.RR()` for propagation checks, logging, and memory lookup. Cleanup now passes the original returned record back to `DeleteRecords`, preserving provider-specific metadata such as `ProviderData`.

## Why

Some libdns providers receive an internal record ID from the provider API when creating a record and store it in `ProviderData`. If CertMagic discards that metadata, cleanup may need to list zone records and rediscover the provider ID before deletion.

Preserving the returned record is more robust:

- Providers with record IDs can delete directly without an extra list call.
- Tokens do not need list permissions solely to rediscover an ID that was already returned during creation.
- Providers without metadata continue to work because the standard DNS data remains available through `Record.RR()`.

This does not make CertMagic depend on provider-specific IDs. It only avoids discarding metadata when a provider supplies it.

## Validation

Added a regression unit test verifying that cleanup receives the original returned record including `ProviderData`.

Tested with:

```sh
go test ./...
```

Fixes #381

### DIFF
--- a/solvers.go
+++ b/solvers.go
@@ -410,7 +410,7 @@ func (m *DNSManager) createRecord(ctx context.Context, dnsName, recordType, reco
 		return zoneRecord{}, fmt.Errorf("expected one record, got %d: %v", len(results), results)
 	}
 
-	return zoneRecord{zone, results[0].RR()}, nil
+	return zoneRecord{zone, results[0]}, nil
 }
 
 // wait blocks until the TXT record created in Present() appears in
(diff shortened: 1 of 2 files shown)


mholt (approved):
Fair enough, didn't realize ProviderData was that useful, but this makes sense! Thanks.

PR #383, opened 2026-05-12
ocsp: add delegated responder authorization regression test

1seal:
ocsp: add delegated responder authorization regression test

## summary

adds a `stapleOCSP` regression test for delegated OCSP responder authorization after #378.

the test covers:
- a CA-issued responder with `ExtKeyUsageOCSPSigning` signs a `Good` response and is stapled
- a same-issuer responder with only `serverAuth` signs a `Good` response for the issued certificate and is rejected before stapling

## tests

```bash
go test -run TestStapleOCSPDelegatedResponderAuthorization -v
go test ./...
```

### DIFF
--- a/ocsp_test.go
+++ b/ocsp_test.go
@@ -4,14 +4,20 @@ import (
 	"bytes"
 	"context"
 	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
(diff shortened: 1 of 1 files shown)


mholt (approved):
Awesome, thanks for the test!

PR #385, opened 2026-05-26
Fix jm.names and jm.activeWorkees leak

u5surf:
In my investigation of the caddy’s issue.
https://github.com/caddyserver/caddy/issues/7366

I found potential resoure leaks in jobmanager.worker if the next.Job causes panic, jm.names is not deleted, activeWorker counter is not decremented.


mholt (approved):
Oh, wow, I don't think I ever saw/noticed that original issue. Thanks for bringing a patch to my attention here. I think this makes sense, let's give it a try. Thank you!

PR #387, opened 2026-06-08
handshake: propagate leader load/obtain outcome to all waiters

Zenexer:
`getCertDuringHandshake` coordinates concurrent handshakes for the same name through `certLoadWaitChans`. If multiple concurrent handshakes occur, they race; the winner loads or obtains a cert, while others wait.

However, the winner's result is only correctly propagated to waiters on the external-cert-manager path (#369). On every other path (on-demand permission denial, a failed obtain, or a successful load whose cert is evicted from the cache before waiters re-check it), losers receive nothing and recursively re-enter the wait queue behind the next winner.

Under sustained handshake load for names that cannot get certificates (e.g. on-demand TLS receiving scanner traffic, or an `ask` endpoint rejecting/throttling), these recursing goroutines accumulate.

This leak was observed in production on caddy 2.11.x, on-demand TLS, ~36 req/s:

* 30k+ goroutines per process at recursion depths up to 48
* Roughly 1.6GB/min of working-set growth

Ideally, ask endpoints should be fast, so this should be a non-issue; in practice, that isn't always the case.

This was debugged and patched largely by Claude, Opus 4.8. I monitored what it was doing and reworked some of its wording, but it seems to have correctly identified the issue without much assistance.


handshake.go
```diff
@@ -334,6 +336,18 @@ func (cfg *Config) getCertDuringHandshake(ctx context.Context, hello *tls.Client
 	// unblock others and clean up when we're done
 	defer func() {
 		certLoadWaitChansMu.Lock()
+		// Propagate this leader's outcome (certificate or error) to all
```
Zenexer: This comment is too verbose for what's basically a three line fix

Zenexer: Trimmed both comment blocks to a few lines; force-pushed (single commit).

mholt (approved):
Thanks, this seems correct, I believe.

PR #396, opened 2026-07-17
Reduce acme challenge log noise

danillouz:
The "looking up info for HTTP challenge" logs can become very noisy (they produced ~2M logs in 24 hours for us)

This PR changes two "expected" cases to log at DEBUG instead of WARN:

- Missing ACME challenges
- Challenge lookups canceled together with their HTTP request

(see the individual commit messages for more detail)

FWIW: these changes have been running in our fork for the last week without any unexpected behavior AFAICS


mholt (approved):
Alright, I think this is an acceptable change, and probably makes the most sense. Thank you! And thanks for your team's sponsorship.

PR #398, opened 2026-08-02
fix: Return last error when retry exhaustion occurs in doWithRetry

vedjaw:
## Summary

When `doWithRetry` exhausted all retry attempts (after `maxRetryDuration` of 30 days), it returned `nil` instead of the last error from the operation. This caused callers to silently treat failed certificate operations as successful.

## Problem

The retry exhaustion path in `doWithRetry` (async.go) returned `nil`:

```go
} else {
    log.Error("final attempt; giving up", ...)
    return nil  // ← BUG: should return the last error
}
```

This triggered a cascading silent-failure chain:

1. **`renewCert`** (`config.go:1037-1044`) calls `doWithRetry` for non-interactive renewals. With the bug, it returned `nil` even after 30 days of continuous failure.
2. **`queueRenewalTask`** (`maintain.go:243-268`) interpreted `nil` as success and called `reloadManagedCertificate` on the old (now expired) certificate.
3. The job manager removed the renewal name from its in-flight set.
4. On the next 10-minute maintenance cycle (`RenewManagedCertificates`), the still-expired cert was re-queued for renewal — but since storage still held the old expired cert, and the renewal would fail again, the cycle repeated indefinitely.

**Net effect:** Certificates could be served expired indefinitely while logs reported "certificate renewed successfully." This is a silent security-relevant failure.

## Root Cause Analysis

`doWithRetry` is designed to retry transient failures with exponential backoff over 30 days. However, if the underlying problem persists (e.g., DNS misconfiguration, CA outage, revoked account), the function must propagate the failure to callers so they can take appropriate action (alert, fall back, etc.). Returning `nil` violated the function's implicit contract: if `f()` never succeeds, the caller must know.

## Fix

One-line change in `async.go`:

```diff
- return nil
+ return err
```

The `err` variable holds the last error returned by `f()`, which is exactly what should be propagated when retries are exhausted.

## Testing

Added comprehensive tests covering all `doWithRetry` exit paths in `async_test.go`:

| Test | Exit Path Verified |
|------|-------------------|
| `TestDoWithRetryReturnsNilOnSuccess` | Success on first attempt → `nil` |
| `TestDoWithRetryReturnsErrorOnErrNoRetry` | `ErrNoRetry` → immediate return, no retries |
| `TestDoWithRetryReturnsErrorOnContextCancellation` | Context cancelled during retry wait → `context.Canceled` |
| `TestDoWithRetryRetriesAndSucceeds` | Transient failures then success → retries occur, returns `nil` |
| `TestDoWithRetryReturnsLastErrorOnExhaustion` | **The regression test for this bug** — verifies `doWithRetry` never returns `nil` when `f()` has failed |

All existing tests continue to pass.

## Checklist

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Tests added/updated
- [x] All tests pass (`go test ./...`)
- [x] Code builds cleanly (`go build ./...`)


mholt (approved):
Thanks, this is indeed probably what is supposed to happen!

PR #399, opened 2026-08-08
storage: clarify Delete vs Load/List/Stat on missing keys

### LINKED ISSUE #310: Clarification on Storage.Delete
## What is your question?
Under what conditions should Delete return an error? [This](https://github.com/caddyserver/certmagic/blob/aad674cda5ba191b4ffce509f05bea5ef8534da9/storage.go#L45) and [this](https://github.com/caddyserver/certmagic/blob/aad674cda5ba191b4ffce509f05bea5ef8534da9/storage.go#L84) seem to contradict each other.

locker95:
The top-level Storage comment said Load/Delete/List/Stat should all return `fs.ErrNotExist` for a missing key, but Delete's method docs say the opposite (only error if the key is still there after the call). FileStorage follows the method docs (`RemoveAll` is fine if the path is already gone).

This just makes the interface comments agree with that.

Fixes #310


mholt:
Thanks, that was confusing indeed.

PR #400, opened 2026-08-20
Fix cache options race during renewal checks

nawtfound404:
## Summary

Fixes a data race in `Config.certNeedsRenewal` when a process-wide CertMagic cache is reused while its options are being updated.

`Cache.SetOptions` already protects writes to `Cache.options` with `optionsMu`, but `certNeedsRenewal` was reading `RenewCheckInterval` without taking the corresponding read lock.

This could race with `TLS.Provision` in Caddy, which calls `SetOptions` when reusing the process-wide cache.

## Changes

- Take `optionsMu.RLock()` once at the start of `certNeedsRenewal`.
- Snapshot `RenewCheckInterval` into a local variable.
- Use the snapshot for the renewal decision and logging.
- Add `TestCertificateNeedsRenewalConcurrentCacheOptions` to exercise concurrent `SetOptions` and `Certificate.NeedsRenewal` calls.

Using one snapshot also ensures that a single renewal decision uses one consistent `RenewCheckInterval`.

## Verification

- `go test -race -run '^TestCertificateNeedsRenewalConcurrentCacheOptions$' -count=20` — passes.
- The same regression test reports a data race with the production fix removed.
- Caddy `TestAvoidDuplicateAutomation` passes with the patched CertMagic under `-race`, 50 repetitions.
- `go test -v -short -race ./...` reaches the new regression test successfully, but the suite currently fails in the existing `TestTryListen` test.
- The `TestTryListen` failure is reproducible against `origin/master` and is unrelated to this change.


mholt (commented):
AI agent review: This duplicates the earlier #397 and applies the same production change. Closing in favor of #397 so review and testing can continue in one place.

PR #403, opened 2026-08-24
make CleanStorage context-aware, suppress cancellation errors

jum:
### Summary
This PR improves `CleanStorage` to properly respect context cancellation across all stages (lock acquisition, OCSP staple deletion, and expired cert cleanup).

### Problem
During server shutdown or configuration reload, if storage cleaning is in progress and the context is canceled, `CleanStorage` and its helper functions would:
1. Continue attempting storage operations until reaching a loop boundary.
2. Return errors (such as `context.Canceled` or transport errors from closed underlying storage backends) which were logged at `Error` level (`"deleting expired certificates staples"`).
3. Attempt to store updated `last_clean.json` even when canceled.

### Solution
- Check `ctx.Err()` before and throughout storage operations in `CleanStorage`, `deleteOldOCSPStaples`, and `deleteExpiredCerts`.
- Suppress error-level logging when an operation aborts due to `context.Canceled`.
- Fix log message typo (`"deleting expired certificates staples"` -> `"deleting expired certificates"`).
- Added unit tests in `maintain_test.go`.

### DIFF
--- a/maintain.go
+++ b/maintain.go
@@ -659,11 +659,16 @@ func CleanStorage(ctx context.Context, storage Storage, opts CleanStorageOptions
 
 	// storage cleaning should be globally exclusive
 	if err := acquireLock(ctx, storage, lockName); err != nil {
+		if errors.Is(err, context.Canceled) || ctx.Err() != nil {
+			return ctx.Err()
+		}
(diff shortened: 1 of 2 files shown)


jum:
As reviewed in caddy issue https://github.com/caddyserver/caddy/issues/7955 this is actually the wrong approach, closing this pr.

PR #406, opened 2026-09-08
handshake: skip building event data nobody will see

### LINKED ISSUE #405: Don't build event data on every handshake when nothing can receive it

`GetCertificateWithContext` emits `tls_get_certificate` as its first statement:

```go
// handshake.go
cfg.emit(ctx, "tls_get_certificate", map[string]any{"client_hello": clientHelloWithoutConn(clientHello)})
```

`emit` returns immediately when `OnEvent` is nil, but Go evaluates arguments first, so the map and the `serializableClientHello` copy are built on every handshake regardless of whether anything can receive them.

I'd like a way for CertMagic to find that out before it builds the data. Something like an optional hook on `Config`:

```go
// Optionally, a function that reports whether the emitter has any
// subscribers for the named event. If not set, CertMagic assumes any
// event may be observed. Must be safe for concurrent use.
HasEventSubscribers func(eventName string) bool
```

consulted through a small helper so the nil `OnEvent` case falls out of it too:

```go
func (cfg *Config) hasSubscribers(eventName string) bool {
	if cfg.OnEvent == nil {
		return false
	}
	if cfg.HasEventSubscribers == nil {
		return true // no way to know; assume yes, as today
	}
	return cfg.HasEventSubscribers(eventName)
}
```

Only `handshake.go` needs the guard. The other twelve `emit` calls are per certificate — caching, issuance, renewal, OCSP — not per handshake, so they can keep calling `emit` directly.

It is on the hottest path CertMagic owns, and the work is discarded a function call later. Measured on master (`31be911`, Apple M2), building that one argument costs **169 ns/op, 672 B/op, 7 allocs/op**, against a `GetCertificate` call that otherwise costs 866 ns/op, 2848 B/op, 24 allocs/op for a cached certificate. That is 20% of the time and 29% of the allocations of the lookup.

The nil-`OnEvent` case alone would not help Caddy, which is presumably the embedder this matters most for: Caddy always installs a hook, because its events app is instantiated whether or not the user configured one. Only the subscriber question helps there, and Caddy can answer it cheaply — its events app already keeps a map of subscriptions per event name and consults it before dispatching, so `HasEventSubscribers` would be a few lines on that side.

The common shape is a server that has no interest in `tls_get_certificate` at all: either it uses no events, or it subscribes to something like `cert_obtained` and pays for `tls_get_certificate` on every handshake anyway.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

None from the embedder side — the cost is incurred before CertMagic calls out, so no `OnEvent` implementation can avoid it.

caddyserver/caddy#7997 makes the dispatch itself cheap on the Caddy side (1241 → 343 ns/op, 3626 → 112 B/op, 28 → 2 allocs/op for a subscriber-less event), but it cannot touch the argument CertMagic has already built. After that change, a handshake on a default Caddy config still pays roughly 169 ns and 672 B here that nothing will ever look at.

## Please link to any relevant issues, pull requests, or other discussions.

- caddyserver/caddy#7996 — where this was found
- caddyserver/caddy#7997 — the Caddy-side change, and the review discussion that led to this split

I have the change and benchmarks ready and will open a PR if the approach looks right.

---

*Assistance disclosure: investigated and written up with Claude Code (Claude Opus 5), which traced the code paths and wrote the benchmarks.*

u5surf:
Closes #405.

## Problem

`GetCertificateWithContext` emits `tls_get_certificate` as its first statement:

```go
cfg.emit(ctx, "tls_get_certificate", map[string]any{"client_hello": clientHelloWithoutConn(clientHello)})
```

`emit` returns immediately when `OnEvent` is nil, but Go evaluates arguments first, so the map and the `serializableClientHello` copy are built on every handshake whether or not anything can receive them.

## Change

An optional `Config.ShouldEmitFunc` lets an embedder say whether an event is worth emitting. It is consulted through a small helper:

```go
func (cfg *Config) shouldEmit(eventName string) bool {
	if cfg.OnEvent == nil {
		return false
	}
	if cfg.ShouldEmitFunc == nil {
		return true // no way to ask; assume it is
	}
	return cfg.ShouldEmitFunc(eventName)
}
```

Four things worth calling out:

**Leaving the field unset keeps today's behavior exactly.** The middle branch is the one every existing user takes: with no way to ask, every event is emitted. Nothing changes for anyone who does not opt in.

**The predicate only comes along with the handler it was written for.** `newWithCache` inherits `ShouldEmitFunc` from `Default` only when it also inherited `OnEvent`; a caller that brings its own handler and no predicate keeps a nil one, rather than picking up an unrelated predicate that knows nothing about it. `TestShouldEmitFuncInheritedWithOnEvent` covers the three combinations.

**The check has to be at the call site.** `emit` already returns early on a nil `OnEvent`, but by the time it is entered its arguments exist. That is the whole problem, and it is why the guard cannot live inside `emit`. The helper's doc comment says so, to keep it from being "simplified" away later.

**Only the handshake is guarded.** The other twelve `emit` calls happen per certificate — caching, issuance, renewal, OCSP — where the data costs less than deciding whether to build it. Sprinkling the guard there would be noise.

## Benchmarks

`GetCertificate` against a cached certificate, with an `OnEvent` handler installed in both cases (Apple M2):

| | ns/op | B/op | allocs/op |
|---|--:|--:|--:|
| `GetCertificateEventSubscribed` | 1000 | 2848 | 24 |
| `GetCertificateEventUnsubscribed` | **777** | **2176** | **17** |

The 672 bytes and 7 allocations are exactly what building the event data costs; measured on its own it is 169 ns/op, 672 B/op, 7 allocs/op. This is not a full handshake — no TLS crypto, no I/O — so it isolates the lookup step.

## Tests

`TestShouldEmit` covers the five states of the two hooks, including that a nil `OnEvent` outweighs a subscriber claim. `TestGetCertificateSkipsUnobservedEvent` checks the behavior through `GetCertificate` for subscribed, unsubscribed, and no-way-to-ask. `TestGetCertificateHonorsEventAbort` pins down that a subscribed handler can still abort a handshake, which is the thing it would be worst to break silently.

## On the Caddy side

Caddy always installs an `OnEvent` hook, so the nil case alone would not help it. Its events app already keeps subscriptions per event name — caddyserver/caddy#7997 uses exactly that to shortcut dispatch — so `ShouldEmitFunc` would be a couple of lines there. I'll send that once this lands and Caddy picks up a release.

---

*Assistance disclosure: I investigated this with Claude Code (Claude Opus 5), which read the code paths, wrote the patch, the benchmarks and the tests, and ran the test suite. I directed the investigation, reviewed the change, and vetted it for correctness.*

### DIFF
--- a/config.go
+++ b/config.go
@@ -70,6 +70,13 @@ type Config struct {
 	// ignore returned errors.
 	OnEvent func(ctx context.Context, event string, data map[string]any) error
 
+	// An optional callback reporting whether OnEvent
+	// has a subscriber for the named event. If unset,
+	// every event is assumed observed. Lets CertMagic
+	// skip building data nobody will see; it is called
+	// concurrently during handshakes, so keep it fast.
+	HasEventSubscribersFunc func(event string) bool
+
 	// DefaultServerName specifies a server name
 	// to use when choosing a certificate if the
 	// ClientHello's ServerName field is empty.
@@ -259,6 +266,9 @@ func newWithCache(certCache *Cache, cfg Config) *Config {
 	}
 	if cfg.OnEvent == nil {
 		cfg.OnEvent = Default.OnEvent
+	}
+	if cfg.HasEventSubscribersFunc == nil {
+		cfg.HasEventSubscribersFunc = Default.HasEventSubscribersFunc
 	}
 	if cfg.KeySource == nil {
 		cfg.KeySource = Default.KeySource
@@ -1352,6 +1362,20 @@ func (cfg *Config) emit(ctx context.Context, eventName string, data map[string]a
 		return nil
 	}
 	return cfg.OnEvent(ctx, eventName, data)
+}
+
+// eventHasSubscriber reports whether emitting the named event could reach
+// anything. Callers use it to avoid building an event's data when nothing
+// will see it; emit() cannot do that itself, since Go evaluates arguments
+// before it is entered. Only worth consulting on hot paths.
+func (cfg *Config) eventHasSubscriber(eventName string) bool {
+	if cfg.OnEvent == nil {
+		return false
+	}
+	if cfg.HasEventSubscribersFunc == nil {
+		return true // no way to ask; assume it is
+	}
+	return cfg.HasEventSubscribersFunc(eventName)
 }
 
 // CertificateSelector is a type which can select a certificate to use given multiple choices.
--- a/config.go
+++ b/config.go
@@ -70,12 +70,15 @@ type Config struct {
 	// ignore returned errors.
 	OnEvent func(ctx context.Context, event string, data map[string]any) error
 
-	// An optional callback reporting whether OnEvent
-	// has a subscriber for the named event. If unset,
-	// every event is assumed observed. Lets CertMagic
-	// skip building data nobody will see; it is called
-	// concurrently during handshakes, so keep it fast.
-	HasEventSubscribersFunc func(event string) bool
+	// An optional callback reporting whether an event
+	// is worth emitting. Returning false skips the
+	// OnEvent call entirely, so it must account for
+	// everything OnEvent does -- logging, metrics --
+	// not only subscribed handlers, and must never
+	// answer no for an event something would observe.
+	// If unset, every event is emitted. Called during
+	// handshakes and concurrently; keep it fast.
+	ShouldEmitFunc func(event string) bool
 
 	// DefaultServerName specifies a server name
 	// to use when choosing a certificate if the
@@ -266,9 +269,12 @@ func newWithCache(certCache *Cache, cfg Config) *Config {
 	}
 	if cfg.OnEvent == nil {
 		cfg.OnEvent = Default.OnEvent
-	}
-	if cfg.HasEventSubscribersFunc == nil {
-		cfg.HasEventSubscribersFunc = Default.HasEventSubscribersFunc
+		// the predicate is written for a specific handler, so it only
+		// comes along with the handler it belongs to; inheriting it onto
+		// a caller's own OnEvent could silence events it wants
+		if cfg.ShouldEmitFunc == nil {
+			cfg.ShouldEmitFunc = Default.ShouldEmitFunc
+		}
 	}
 	if cfg.KeySource == nil {
 		cfg.KeySource = Default.KeySource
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -70,6 +70,13 @@ type Config struct {
 	// ignore returned errors.
 	OnEvent func(ctx context.Context, event string, data map[string]any) error
 
+	// An optional callback reporting whether OnEvent
+	// has a subscriber for the named event. If unset,
+	// every event is assumed observed. Lets CertMagic
+	// skip building data nobody will see; it is called
+	// concurrently during handshakes, so keep it fast.
+	HasEventSubscribersFunc func(event string) bool
```
steadytao: Returning false suppresses `OnEvent` entirely, not merely subscriber dispatch. Please either rename this around event emission or document that it must account for every observer, remain concurrency-safe and never return a false negative. Our debug logging is an observer even when no handler is subscribed.

config.go
```diff
@@ -260,6 +267,9 @@ func newWithCache(certCache *Cache, cfg Config) *Config {
 	if cfg.OnEvent == nil {
 		cfg.OnEvent = Default.OnEvent
 	}
+	if cfg.HasEventSubscribersFunc == nil {
+		cfg.HasEventSubscribersFunc = Default.HasEventSubscribersFunc
+	}
```
steadytao, line 278: These callbacks must be defaulted as a pair. If the caller supplies `OnEvent` but leaves this nil, the unrelated predicate from `Default` can suppress the callers handler. Only inherit the default predicate when `OnEvent` was also inherited and add a regression test for this case.

steadytao (changes requested):

config.go
u5surf: Renamed to `ShouldEmitFunc` and rewrote the doc around emission rather than subscribers: returning false skips the `OnEvent` call entirely, so it has to account for everything `OnEvent` does — not only subscribed handlers — must be safe for concurrent use, and must never answer no for an event something would have observed.

Your debug-logging example is exactly the trap. Caddy writes a line per event whether or not anything is subscribed, so a predicate answering the literal "do I have subscribers?" question would have silenced its own logging.

config.go
u5surf: Fixed — the predicate is now inherited only alongside the handler it was written for:

```go
if cfg.OnEvent == nil {
	cfg.OnEvent = Default.OnEvent
	if cfg.ShouldEmitFunc == nil {
		cfg.ShouldEmitFunc = Default.ShouldEmitFunc
	}
}
```

A caller that supplies its own `OnEvent` and no predicate now keeps a nil one and has its events emitted, instead of inheriting a predicate that knows nothing about it. `TestShouldEmitFuncInheritedWithOnEvent` covers the three combinations: own handler, inherited handler, and own predicate with an inherited handler.

steadytao (approved):
Thank you. It does lgtm but ill let Matt or Francis touch seeing as I am not too familiar with this codebase -- @mholt @francislavoie

francislavoie (approved):

